PR #12, opened 2018-05-07
deploy: set the default storage class to DigitalOcean

### LINKED ISSUE #11: Set the storage class default in the deploy scripts

fatih:
This aligns with our plugin and also makes it easier to deploy
PersistentVolumeClaims because users don't have to specify a
`storageClassName` anymore.

closes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,11 @@ kubectl create -f csi-nodeplugin-rbac.yaml
 kubectl create -f csi-nodeplugin-do.yaml
 ```
 
+A new storage class will be created with the name `do-block-storage` which is
+responsible for dynamic provisioning. This is set to "default" for dynamic
+provisioning. If you're using multiple storage classes you might want to remove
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,11 @@ kubectl create -f csi-nodeplugin-rbac.yaml
 kubectl create -f csi-nodeplugin-do.yaml
 ```
 
+A new storage class will be created with the name `do-block-storage` which is
```
aybabtme: `This is set to default` -> `This is set by default` or `This is true by default`?

aybabtme:
:shipit:

README.md
fatih: If you install a storage class, it's not set to `default` setting. If you mean `set by default`, the question boils down on `what` is set to default. If I change then it should be something like `Thi is set by default to "default"` which is not what we want?

fatih: Changed it to `This is set to "default" for dynamic provisioning.`

PR #15, opened 2018-05-08
driver: log method errors

### LINKED ISSUE #2: CSI: Add ErrorLog interceptor for the gRPC servers
We have multiple methods in the plugin that might return an error. We should add an interceptor that logs the return error of all functions automatically. I've didn't looked into it, but I believe there might be something like this.

fatih:
Instead of logging return errors in all methods, we only log it once via
the interceptor

example:

image

closes #2


aybabtme:
:shipit:

PR #17, opened 2018-05-09
controller: make remaining functions idempotent

### LINKED ISSUE #7: Make sure functions are idempotent

fatih:
closes #7


driver/node.go
```diff
@@ -116,11 +131,18 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
...
 	if err != nil {
 		return nil, err
 	}
 
+	if mounted {
+		err := d.mounter.Unmount(req.StagingTargetPath)
+		if err != nil {
+			return nil, err
+		}
+	}
```
nanzhong: For consistency in logging, should we also log that the device is already not mounted here like for the other actions?

nanzhong:
lgtm, one tiny comment about logging, not a blocker.

driver/node.go
fatih: You're right. I've missed this part. Fixed and force pushed it 👍

PR #21, opened 2018-05-11
Change secret name to digitalocean

### LINKED ISSUE #20: Align k8s digitalocean token secret name with cloud-controller-manager
There is a difference in the name between how digitalocean/cloud-controller-manager calls the secret that holds the k8s digitalocean token, and the way the secret is named here. Leading to two secrets with different names in k8s.

Is it possible to align these ?

fatih:
This means that users that already have created a token for
cloud-controller-manager can use the same token for the CSI plugin

- [x] checkout namespace compatibility

Fixes #20

### DIFF
--- a/deploy/kubernetes/releases/csi-digitalocean-NOT_RELEASED_v0.0.2.yaml
+++ b/deploy/kubernetes/releases/csi-digitalocean-NOT_RELEASED_v0.0.2.yaml
@@ -1,0 +1,354 @@
+# Copyright 2018 DigitalOcean
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+ 
+# Configuration to deploy release version of the CSI DigitalOcean
+# plugin (https://github.com/digitalocean/csi-digitalocean) compatible with
+# Kubernetes >=v1.10
+#
+# example usage: kubectl create -f <this_file>
+
+---
+kind: StorageClass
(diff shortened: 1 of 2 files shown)


fatih:
@andrewsykim can you please comment on the namespace compatibility. I see that CCM is using `kube-system`, but we don't do it for CSI. I believe therefore, even if I change the name, it's not going to work out because the CCM secret lives in a separate namespace. The CSI secret lives in the `default` namespace. Happy to change it for our own case, just wonder the reason why CCM uses `kube-system`.

deploy/kubernetes/releases/csi-digitalocean-NOT_RELEASED_v0.0.2.yaml
```diff
@@ -0,0 +1,354 @@
...
+            - name: ADDRESS
+              value: /csi/csi.sock
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /csi/
+              # TODO(arslan): the registar is not implemented yet
```
aybabtme: `registrar`

andrewsykim:
CCM uses `kube-system` since it is technically a component of the kubernetes control plane. Like you said I think the best option going forward (at least for now) is just to duplicate the secret for this use case.

fatih:
I see. I'm not sure why the CSI plugin should live in `kube-system`. That namespace ism meant for services that are [created by Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/):

> The namespace for objects created by the Kubernetes system

Do you think it's better we change the default namespace for CCM to `default`?

aybabtme:
Seems to me that if `CCM` is a `kube-system` then the case is even stronger for `CSI` to be a `kube-system`?

fatih:
@aybabtme true that. Moving it to `kube-system`. Will update the PR now

fatih:
@aybabtme @andrewsykim moved to `kube-system` and fixed review comments. PTAL.

fatih:
@lxfontes thanks for catching these. Fixed it.

lxfontes:
lgtm

cagedmantis:
lgtm

PR #22, opened 2018-05-11
Add method names for each log entry

### LINKED ISSUE #16: Add method name to the log fields
It's better to have a `method:foo` in our logs instead of logging `foo is called`. This makes it easy to follow logs for a given method

fatih:
closes #16


aybabtme:
:shipit:

PR #23, opened 2018-05-11
Add Changelog.md

fatih:
We started the alpha with the patch version (0.0.1). Hence we should start the beta with (0.1.0) going forward so we have space for patches if needed.


aybabtme:
:shipit:

Issue #33, opened 2018-05-26
Panic when deleting


metalmatze:
Hey, I already have 180 restarts on my pod `csi-provisioner-doplugin-0`.

Here are the logs:
```
time="2018-05-26T16:01:21Z" level=info msg="get plugin info called" method=get_plugin_info node_id=94784XXX region=fra1 response="name:\"com.digitalocean.csi.dobs\" vendor_version:\"0.1.0\" "
time="2018-05-26T16:01:21Z" level=info msg="delete volume called" method=delete_volume node_id=94784XXX region=fra1 volume_id=252ec68a-4664-11e8-9dc3-0242ac110XXX
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7c8e35]

goroutine 32 [running]:
github.com/digitalocean/csi-digitalocean/driver.(*Driver).DeleteVolume(0xc420094900, 0x94f040, 0xc4201b24e0, 0xc420170760, 0xc420094900, 0x0, 0x0)
	/Users/fatih/go/src/github.com/digitalocean/csi-digitalocean/driver/controller.go:152 +0x335
github.com/digitalocean/csi-digitalocean/vendor/github.com/container-storage-interface/spec/lib/go/csi/v0._Controller_DeleteVolume_Handler.func1(0x94f040, 0xc4201b24e0, 0x88a220, 0xc420170760, 0x7fbaabe886c8, 0x0, 0x8a7ac0, 0xc420170740)
	/Users/fatih/go/src/github.com/digitalocean/csi-digitalocean/vendor/github.com/container-storage-interface/spec/lib/go/csi/v0/csi.pb.go:2148 +0x86
github.com/digitalocean/csi-digitalocean/driver.(*Driver).Run.func1(0x94f040, 0xc4201b24e0, 0x88a220, 0xc420170760, 0xc420170780, 0xc4201707a0, 0x853ca0, 0xbe0bc8, 0x8d0260, 0xc4204112c0)
	/Users/fatih/go/src/github.com/digitalocean/csi-digitalocean/driver/driver.go:129 +0x78
[...]

PR #34, opened 2018-05-26
Avoid panicking on errors without response object

nussjustin:
In some cases when an action fails the controller looks at the response status code to determine if the error can be ignored.

The problem is that for some error cases the response itself can be nil which then leads to a panic when accessing the StatusCode field.

I took a quick look at the code and found 3 places where this could happen. This PR updates the code to check for a non-nil response first in all 3 places, avoiding the panic. 

Updates #33


fatih:
Thanks @nussjustin for the fix

PR #46, opened 2018-08-01
driver: check if the mounts are propagated

fatih:
Make sure mount propagation is enabled. We can check it by using
`findmnt` with the `PROPAGATION` label.

Also fix an issue to check the mounted path for the target path where
checking against the staging target path won't work as findmnt only
works on devices

### DIFF
--- a/driver/mounter.go
+++ b/driver/mounter.go
@@ -17,12 +17,24 @@ limitations under the License.
 package driver
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
 )
+
+type findmntResponse struct {
+	FileSystems []fileSystem `json:"filesystems"`
+}
+
+type fileSystem struct {
+	Target      string `json:"target"`
+	Propagation string `json:"propagation"`
+	FsType      string `json:"fstype"`
+	Options     string `json:"options"`
+}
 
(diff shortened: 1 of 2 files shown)


driver/mounter.go
```diff
@@ -194,9 +202,28 @@ func (m *mounter) IsMounted(source, target string) (bool, error) {
...
+	targetFound := false
+	for _, fs := range resp.FileSystems {
+		// check if the mount is propagated correctly. It should be set to shared.
+		if fs.Propagation != "shared" {
+			return true, fmt.Errorf("mount propagation for target %q is not enabled", target)
+		}
+
+		// the mountpoint should match as well
+		if fs.Target == target {
+			targetFound = true
```
lxfontes: `return true, nil` and get rid of `targetFound`

lxfontes:
lgtm

driver/mounter.go
fatih: Good call, but we can't do it because there might be two mount points. If the first one is `shared` and it fits the target, if we early return in this case, we might miss the second one (which could be `private` for example). However I removed this piece:

```
	if targetFound {
		return true, nil
	}

	return false, nil
```

to 

```
	return targetFound, nil
```

lxfontes: ❤️

lxfontes:
🚀

PR #50, opened 2018-08-03
driver: pass the correct source to IsMounted()

fatih:
We shouldn't pass an empty source path to IsMounted. Also make sure it's
never empty.


lxfontes:
lgtm

PR #59, opened 2018-08-20
driver: remove desc check

fatih:
The description might change, which would avoid making the check
invalid. Avoid doing it.


fatih:
cc @lxfontes

lxfontes:
lgtm on 🍏 
it's not that it might change, it might have been created by a different system ( ex: control panel or manually via API )

PR #60, opened 2018-08-21
driver: handle non existing volumes and nodes

fatih:
According to the spec we should return a `NOT_FOUND` in various method.
We always assume the volume or node exist, however this might not the
case.

For example the user might delete the volume externally from the
UI or make a HTTP call to the DigitalOcean API to delete the volume.
Actions like this are outside the Kubernetes system, so the plugin needs
to handle these edge cases by signaling back that the resources don't
exist.

We return `NOT_FOUND` for any creation actions, such as creating a
volume, attaching a volume, formatting, etc.. Because we really need to
return an error as there is no way to recover from this

We don't return `NOT_FOUND` for any destroy actions, such as deleting,
detaching, unmounting, etc.. Because in all these cases it doesn't
matter if the volume doesn't exist. There is not much to do and assuming
it's "done" makes the overall system more stable

### DIFF
--- a/driver/node.go
+++ b/driver/node.go
@@ -64,6 +64,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
 			return nil, status.Errorf(codes.NotFound, "volume %q not found", req.VolumeId)
 		}
+		return nil, err
 	}
 
 	source := getDiskSource(vol.Name)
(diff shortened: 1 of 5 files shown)


driver/node.go
```diff
@@ -248,27 +258,15 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
...
-
-	source := getDiskSource(vol.Name)
-	mounted, err := d.mounter.IsMounted(source, req.TargetPath)
-	if err != nil {
+	ll.Info("unmounting the target path")
+	err := d.mounter.Unmount(req.TargetPath)
+	// Invalid argument is returned by the `mount` command in this container
+	// image when a path is already umounted. We only return an error for
+	// anything else
+	if err != nil && !strings.Contains(err.Error(), "Invalid argument") {
```
nanzhong: Mhmm this string comparison feels like it’s a little fragile, is there another way to determine the error case?

nanzhong:
lgtm, one question about checking the error.

fatih:
The `Invalid argument` check will be removed and replaces. Please don't take that into account.

driver/node.go
fatih: Good call, as I said I'll use `IsMounted` that doesn't rely on the `Source` so it's a safe bet

lxfontes:
🚀

PR #61, opened 2018-08-22
driver: make sure to handle concurrent calls to the droplet

### LINKED ISSUE #32: Intermittent MountVolume.MountDevice errors on Pod creation
I've set up Rancher 2.0 cluster on DO to test CSI-DO. At first attempt i followed README.md and succeeded with example app. However after trying to do my own stuff I've started consistently getting Pod creation errors. To rule out the unknowns I've wiped all my stuff and returned back to example app and confirmed that i'm consistently getting this result:
```
Events:
  Type     Reason                  Age                From                     Message
  ----     ------                  ----               ----                     -------
  Warning  FailedScheduling        21s (x4 over 22s)  default-scheduler        pod has unbound PersistentVolumeClaims
  Normal   Scheduled               19s                default-scheduler        Successfully assigned my-csi-app to node-1
  Normal   SuccessfulMountVolume   19s                kubelet, node-1          MountVolume.SetUp succeeded for volume "default-token-jw8hg"
  Normal   SuccessfulAttachVolume  15s                attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-4b1897fb-5f66-11e8-8f45-c2743f7bbff2"
  Warning  FailedMount             5s (x4 over 10s)   kubelet, node-1          MountVolume.MountDevice failed for volume "pvc-4b1897fb-5f66-11e8-8f45-c2743f7bbff2" : rpc error: code = Internal desc = formatting disk failed: exit status 1 cmd: 'mkfs.ext4 -F /dev/disk/by-id/scsi-0DO_Volume_pvc-4b1897fb-5f66-11e8-8f45-c2743f7bbff2' output: "mke2fs 1.43.7 (16-Oct-2017)\nThe file /dev/disk/by-id/scsi-0DO_Volume_pvc-4b1897fb-5f66-11e8-8f45-c2743f7bbff2 does not exist and no size was specified.\n"
```

However just after waiting for couple hours problem is gone. I wonder if it was Block Storage degradation I've just happened to witness or this is something related to CSI-DO. Unfortunatelly I've wiped out that cluster and after setting up new one example app deploys just fine. I will provide additional info you might need if I witness this problem again.

fatih:
The DigitalOcean API doesn't process actions for a given resource in
async manner. Calls needs to be done sequentially. This means, if there
is an ongoing process, say an "attach" process, another "attach" or
"detach" will fail.

We return a `ABORTED` error, which makes sure the csi-attacher to retry
again in the next backoff tick (see:
https://github.com/kubernetes-csi/external-attacher/blob/5a4cc3b1af15966cdb9b8ca1af6ab3adade877b0/pkg/controller/csi_handler.go#L307
and
https://github.com/kubernetes-csi/external-attacher/blob/5a4cc3b1af15966cdb9b8ca1af6ab3adade877b0/pkg/connection/connection.go#L223-L227
)

fixes #32

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -202,12 +202,24 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 	action, resp, err := d.doClient.StorageActions.Attach(ctx, req.VolumeId, dropletID)
 	if err != nil {
 		// don't do anything if attached
-		if (resp != nil && resp.StatusCode == http.StatusUnprocessableEntity) || strings.Contains(err.Error(), "This volume is already attached") {
-			ll.WithFields(logrus.Fields{
-				"error": err,
-				"resp":  resp,
-			}).Warn("assuming volume is attached already")
-			return &csi.ControllerPublishVolumeResponse{}, nil
+		if resp != nil && resp.StatusCode == http.StatusUnprocessableEntity {
+			if strings.Contains(err.Error(), "This volume is already attached") {
+				ll.WithFields(logrus.Fields{
+					"error": err,
+					"resp":  resp,
+				}).Warn("assuming volume is attached already")
+				return &csi.ControllerPublishVolumeResponse{}, nil
+			}
+
+			if strings.Contains(err.Error(), "Droplet already has a pending event") {
+				ll.WithFields(logrus.Fields{
+					"error": err,
+					"resp":  resp,
+				}).Warn("droplet is not able to detach the volume")
+				// sending an abort makes sure the csi-attacher retries with the next backoff tick
+				return nil, status.Errorf(codes.Aborted, "volume %q couldn't be attached. droplet %d is in process of another action",
+					req.VolumeId, dropletID)
+			}
 		}
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


nanzhong:
lgtm

driver/controller.go
```diff
@@ -202,12 +202,24 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 	action, resp, err := d.doClient.StorageActions.Attach(ctx, req.VolumeId, dropletID)
 	if err != nil {
 		// don't do anything if attached
-		if (resp != nil && resp.StatusCode == http.StatusUnprocessableEntity) || strings.Contains(err.Error(), "This volume is already attached") {
-			ll.WithFields(logrus.Fields{
-				"error": err,
-				"resp":  resp,
-			}).Warn("assuming volume is attached already")
-			return &csi.ControllerPublishVolumeResponse{}, nil
+		if resp != nil && resp.StatusCode == http.StatusUnprocessableEntity {
```
peterver: Looks like there's some code duplication in regards to the checks here ?

driver/controller.go
```diff
@@ -202,12 +202,24 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
+			if strings.Contains(err.Error(), "This volume is already attached") {
+				ll.WithFields(logrus.Fields{
+					"error": err,
+					"resp":  resp,
+				}).Warn("assuming volume is attached already")
+				return &csi.ControllerPublishVolumeResponse{}, nil
+			}
+
+			if strings.Contains(err.Error(), "Droplet already has a pending event") {
```
peterver: Doesn't digitalocean return a specific error code for this? What if the string itself changes

peterver: such as for example the rate limit return message has the following as body : 

```
{
        id: "too_many_requests",
        message: "API Rate limit exceeded."
}
```

for which the id would be a prime use case here :)

driver/controller.go
fatih: Peter how exactly? I don't see any duplication. Can you give a more concrete example?

driver/controller.go
fatih: As for now, this is the only way we can deal with it. But we'll have some internal todo's to make this better :)

fatih:
I'll merge this as I want to continue on debugging and also release a new version so we can make sure people are up to date. Follow up fixes will be opened.

driver/controller.go
fatih: Oh I see, a little copy is ok I guess :) I'll refactor if needed though.

PR #70, opened 2018-08-27
gh: add instructions to add logs to newly created issues

fatih:
Without logs there is not much we can do


lxfontes:
lgtm

💡  check if it makes sense to request `kubectl cluster-info dump` instead. it will carry all this info and more.

fatih:
@lxfontes this is much better, has tons of information. A little too much, but it's not something bad to have. Changing the text 👍

PR #80, opened 2018-09-13
Integration tests

### LINKED ISSUE #18: Add e2e tests
We should have some tests that spin up a droplet, installs the driver via cloudinit and then uses https://github.com/rexray/gocsi/tree/master/csc to test out the tests in real.

fatih:
This adds the initial framework for running integration tests for Kubernetes. I opened it because the PR get's bigger and bigger. To run the integration test, point to a valid cluster configuration file with the Plugin pre-installed:  

```
$ KUBECONFIG=$(pwd)/kubeconfig make test-integration
```

This will make sure to run the individual test cases against this cluster. Things that need improvements, will be added in upcoming PR's:

* Check CSI and K8S version before running the tests to make sure we're running against the same set of dependencies
* Add and diversify test cases (currently it's just deploying a PVC and Pod). Create more challenging, real world tests

Closes #18


test/kubernetes/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	fmt.Println("Finished!")
+}
+
+func setup() error {
+	var kubeconfig string
+	flag.StringVar(&kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file")
+	flag.Parse()
+
+	if kubeconfig == "" {
+		kubeconfig = os.Getenv(clientcmd.RecommendedConfigPathEnvVar)
```
lxfontes: i don't think you need all this.

```
loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
// if you want to change the loading rules (which files in which order), you can do so here

configOverrides := &clientcmd.ConfigOverrides{}
// if you want to change override values or bind them to flags, there are methods to help you

kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
config, err := kubeConfig.ClientConfig()
if err != nil {
	// Do something
}
```

it will read kubeconfig from the usual sources OR from `KUBECONFIG`

test/kubernetes/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil
+}
+
+func teardown() error {
+	// delete all test resources
+	err := client.CoreV1().Namespaces().Delete(namespace, nil)
+	if err != nil && !(kubeerrors.IsNotFound(err) || kubeerrors.IsAlreadyExists(err)) {
+		return err
+	}
+
```
lxfontes: 💭 do we need to wait for complete namespace deletion?
rationale: if another test tries to reuse a namespace it will fail as the namespace is still being deleted

test/kubernetes/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+		"pods", namespace, fields.Everything())
+	_, controller := cache.NewInformer(watchlist, &v1.Pod{}, time.Second*1,
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(o, n interface{}) {
+				pod := n.(*v1.Pod)
+				if name != pod.Name {
+					return
+				}
+
+				if pod.Status.Phase == v1.PodRunning {
```
lxfontes: should also stop on `PodFailed` and `PodSucceeded`

lxfontes:
@fatih feel free to merge this in and fix-forward
it doesn't block releases atm so it is all good

test/kubernetes/integration_test.go
fatih: Thanks for the suggestion. Changed it with the snippet above.

test/kubernetes/integration_test.go
fatih: @lxfontes yes we need, if you run consecutive runs, it'll fail (I've seen it several times). I'm actually aware of this, this check is only temporary, I wanted to add more bullet prof checks later in the process. I'll make sure to fix it indeed 👍

PR #82, opened 2018-09-18
Test improvements

fatih:


test/kubernetes/integration_test.go
```diff
@@ -30,15 +32,13 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	fmt.Println("==> Setting up tests")
```
lxfontes: 👏

test/kubernetes/integration_test.go
```diff
@@ -108,18 +108,236 @@ func TestPod_Single_Volume(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	pods, err := client.CoreV1().Pods(namespace).
+		List(metav1.ListOptions{LabelSelector: selector.String()})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pod := pods.Items[0]
```
lxfontes: would we ever need to check for this boundary ?
might be safe, just double checking

lxfontes:
lgtm

test/kubernetes/integration_test.go
```diff
@@ -108,18 +108,236 @@ func TestPod_Single_Volume(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	t.Log("Creating deployment")
+	_, err := client.AppsV1().Deployments(namespace).Create(dep)
```
nanzhong: Any specific reason we are creating a deployment here for the test vs just a pod?

fatih: Pod is another test (See above in the file). This is Deployment + PVC, we already test for Pod + PVC.

test/kubernetes/integration_test.go
fatih: Makes sense, I'll test it 👍

test/kubernetes/integration_test.go
fatih: @nanzhong if you're asking for the difference, pod doesn't get rescheduled, deployment will reschedule pod if pod get killed. We're going to add more test cases, such as removing a node and observe how a pod is getting scheduled. This will not be possible for pod based tests.

PR #87, opened 2018-10-03
csi: document how to use existing volumes

### LINKED ISSUE #85: Connect existing block storage volumes?
Hi there,

is it possible to use *csi-digitalocean* to connect a kubernetes pod to an existing block storage volume?

I looked at the docs, examples and the issues, but I couldn't find anything related to this.

fatih:
* Add Readme with examples on how to use existing volumes
* Add an option to prevent formatting an existing volume (needs more
testing)

KNOWN ISSUES:

* Deleting a `PVC` will delete the volume if the reclaim policy is set
to `Delete`, however the `PV` will be not deleted. This seems to be a
bug in K8S
* Using an existing, attached volume might cause confusion if the POD is
scheduled to a different node. It's advised that user only use existing
**detached** volumes

closes #85

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -180,6 +180,10 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 	if err != nil {
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
 			// we assume it's deleted already for idempotency
+			ll.WithFields(logrus.Fields{
+				"error": err,
+				"resp":  resp,
+			}).Warn("assuming volume is deleted already")
 			return &csi.DeleteVolumeResponse{}, nil
 		}
 		return nil, err
@@ -227,7 +231,7 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 	ll.Info("controller publish volume called")
 
 	// check if volume exist before trying to attach it
-	_, resp, err := d.doClient.Storage.GetVolume(ctx, req.VolumeId)
+	vol, resp, err := d.doClient.Storage.GetVolume(ctx, req.VolumeId)
 	if err != nil {
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
 			return nil, status.Errorf(codes.NotFound, "volume %q not found", req.VolumeId)
@@ -244,6 +248,22 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 		return nil, err
 	}
 
+	attachedID := 0
+	for _, id := range vol.DropletIDs {
+		attachedID = id
+		if id == dropletID {
+			ll.Info("volume is already attached")
+			return &csi.ControllerPublishVolumeResponse{}, nil
+		}
+	}
+
+	// droplet is attached to a different node, return an error
+	if attachedID != 0 {
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"volume is attached to a wrong droplet(%q), dettach the volume to fix it", attachedID)
+	}
+
+	// attach the volume to the correct node
 	action, resp, err := d.doClient.StorageActions.Attach(ctx, req.VolumeId, dropletID)
 	if err != nil {
 		// don't do anything if attached
--- a/examples/kubernetes/pod-single-existing-volume/README.md
+++ b/examples/kubernetes/pod-single-existing-volume/README.md
@@ -1,0 +1,147 @@
+# Use an existing volume
+
+Below you will find the instruction on how to use an existing DigitalOcean Block
+Storage with your Kubernetes cluster.
+
+## Known issues
+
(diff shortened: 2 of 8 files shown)


examples/kubernetes/pod-single-existing-volume/README.md
```diff
@@ -0,0 +1,146 @@
+# Use an existing volume
+
+Below you find the instruction on how to use an existing DigitalOcean Block
```
aybabtme: `Below you will find`

examples/kubernetes/pod-single-existing-volume/README.md
```diff
@@ -0,0 +1,146 @@
...
+      com.digitalocean.csi/format: "true"
+```
+
+Couple of things to note,
+
+* `volumeHandle` is the volume ID you want to reuse. Make sure it matches exactly the volume you're targeting. You can list the ID's of your volumes via doctl: `doctl compute volume list`
+* `volumeAttributes` has a special, csi-digitalocean specific annotation called `com.digitalocean.csi/format`. If you add this key, the CSI plugin make sure to **not format** the volume. If you don't add this, it'll be formatted.
+* `storage` make sure it's set to the same storage size as your existing DigitalOcean Block Storage volume.
+
+Once everything is set, create this file called `pv.yaml` and deploy it:
```
aybabtme: `create a file with this content, naming it `pv.yaml`, and deploy it`

aybabtme:
lgtm

lxfontes:
@fatih just to confirm on
> Deleting a PVC will delete the volume if the reclaim policy is set
to Delete, however the PV will be not deleted. This seems to be a
bug in K8S

The expected behaviour here is `reclaimPolicy: Delete` to delete pvc & pv and that is not working, correct? If so, this explains an internal tester reporting it. otherwise we need to 🔍

driver/controller.go
```diff
@@ -177,11 +177,90 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
...
+		return &csi.DeleteVolumeResponse{}, nil
+	}
+
+	// there are no ways to intepret the error, return it
+	if resp == nil {
+		return nil, err
+	}
+
+	// we assume it's deleted already for idempotency
+	if resp.StatusCode == http.StatusNotFound {
```
lxfontes: log here

lxfontes:
lgtm
minor comments

fatih:
@lxfontes 

> The expected behaviour here is reclaimPolicy: Delete to delete pvc & pv and that is not working, correct? If so, this explains an internal tester reporting it. otherwise we need to 🔍

For dynamic provisioning, such as when you set a storage class (Which is the case if you deployr our CSI plugin or in DigitalOcean Kubernetes), in this case, the `reclaimPolicy` is automatically set to `Delete`.  If you created a volume dynamically, (by just using a `PVC` and `POD), in this case, removing a `PVC` removes the `PV` as well. 

However, in this case, if you manually create a `PV` and then create a `PVC`, you decide to delete the `PVC`, the `PV` will be not removed. The backed volume will be deleted though. The issue is, somehow the finalizer is not removed from the `PV` resource, so it's get stuck. If you do a `kubectl edit pv ...` and remove the finalizer (which is added by the storage controller), it's get deleted immediately. 

I think this is a bug on K8S. I'll try to reproduce and open a bug issue if I can find something

driver/controller.go
fatih: This also happens if the plugin receives two or more delete calls. The first one deletes it, the second one gets a `StatusNotFound`.

driver/controller.go
```diff
@@ -323,6 +248,22 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
+		if id == dropletID {
+			ll.Info("volume is already attached")
+			return &csi.ControllerPublishVolumeResponse{}, nil
+		}
+	}
+
+	// droplet is attached to a different node, return an error
+	if attachedID != 0 {
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"volume is attached to a wrong droplet(%q), dettach the volume to fix it", attachedID)
```
aybabtme: `volume is attached to the wrong droplet(%q), detach the volume to fix it`

examples/kubernetes/pod-single-existing-volume/README.md
```diff
@@ -50,7 +51,7 @@ Couple of things to note,
 * `volumeAttributes` has a special, csi-digitalocean specific annotation called `com.digitalocean.csi/format`. If you add this key, the CSI plugin make sure to **not format** the volume. If you don't add this, it'll be formatted.
 * `storage` make sure it's set to the same storage size as your existing DigitalOcean Block Storage volume.
 
-Once everything is set, create this file called `pv.yaml` and deploy it:
+Create a file with this content, naming it `pv.yaml` and deploy it:
```
aybabtme: `deploying`

aybabtme:
lgtm

PR #94, opened 2018-10-29
Switch to Go modules.

### LINKED ISSUE #78: Switch from dep to Go modules
With Go v1.11, Go modules is now an experimental mode integrated into the Go toolchain. It allows native dependency management without the need of an additional tool. Also Go modules allows us to develop the plugin without the need of a GOPATH, which makes it easier for first-time contributors or Go users. In overall it has many benefits compared to other tools.

timoreimann:
We keep the vendor folder but repopulate it through Go modules.

Also bump minimum required Go version to 1.11 for dependency management reasons.

Fixes #78


timoreimann:
@fatih I wasn't exactly sure if we wanted to keep building for Go 1.10 in Travis or drop support since module-based dependency management forces us to go with 1.11 to my understanding.

Happy to adjust, of course.

timoreimann:
Pushed another commit with the result of `go mod tidy` executed. Keeping it separate for now to see the diff more easily, happy to squash latter.

fatih:
Thanks @timoreimann. I tested this on a folder outside GOPATH and everything works fine (including unit and integration tests).

PR #96, opened 2018-10-31
csi: fix driver name

fatih:
Driver name should match the StorageClass, CSIDriver and the name inside
our driver, which is returned via GetNodeInfo().

I'll release a new version once that is merged


nanzhong:
lgtm

PR #99, opened 2018-11-09
csi: use publish_info for storing and accessing the volume name

joonas:
After tracing through a bunch of the CSI code, I came to the conclusion that this could be a viable strategy for doing away with the need for the nodes to call the API to fetch the volume name.

Super open to feedback here, I've verified that this does indeed work by running the integration test suite against a cluster with my custom build deployed on it for CSI.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -41,6 +41,8 @@ const (
 )
 
 const (
+	PublishInfoVolumeName = "com.digitalocean.csi/volume-name"
+
 	defaultVolumeSizeInGB = 16 * GB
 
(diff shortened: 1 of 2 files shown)


fatih:
This looks good @joonas. Seems like I totally forgot that the response has something like this. There are couple of improvements we can make, such as using `"dobs.csi.digitalocean.com"` as a prefix (CSI uses a forward domain notation now instead reverse) and adding a comment to the `PublishInfoVolumeName`. I'll merge and make the changes myself, and also update the changelog with the changes. I make sure to test it as well.

PR #101, opened 2018-11-10
CreateVolume RPC refinements

joonas:
After looking through the `CreateVolume` code, I thought that there was a few improvements I could make to it, so here they are.. Let me know if you'd prefer me to break this up or drop some of these.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -18,7 +18,6 @@ package driver
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -45,9 +44,17 @@ const (
 	// `ControllerPublishVolume` to `NodeStageVolume or `NodePublishVolume`
 	PublishInfoVolumeName = DriverName + "/volume-name"
 
-	// defaultVolumeSizeInGB is used when the user didn't defined a correct
-	// storage size or if the size is not satisfised
-	defaultVolumeSizeInGB = 16 * GB
+	// minimumVolumeSizeInBytes is used to validate that the user is not trying
+	// to create a volume that is smaller than what we support
+	minimumVolumeSizeInBytes int64 = 1 * GB
+
+	// maximumVolumeSizeInBytes is used to validate that the user is not trying
+	// to create a volume that is larger than what we support
+	maximumVolumeSizeInBytes int64 = 16 * TB
+
+	// defaultVolumeSizeInBytes is used when the user did not provide a size or
+	// the size they provided did not satisfy our requirements
+	defaultVolumeSizeInBytes int64 = 16 * GB
 
 	// createdByDO is used to tag volumes that are created by this CSI plugin
 	createdByDO = "Created by DigitalOcean CSI driver"
@@ -93,7 +100,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 
 	size, err := extractStorage(req.CapacityRange)
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Error(codes.InvalidArgument, "invalid capacity range: %v", err)
 	}
 
 	volumeName := req.Name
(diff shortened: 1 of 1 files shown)


driver/controller.go
```diff
@@ -603,31 +609,65 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
-		return defaultVolumeSizeInGB, nil
+		return defaultVolumeSizeInBytes, nil
 	}
 
-	if capRange.RequiredBytes == 0 && capRange.LimitBytes == 0 {
-		return defaultVolumeSizeInGB, nil
+	requiredBytes := capRange.GetRequiredBytes()
+	requiredSet := 0 < requiredBytes
+	limitBytes := capRange.GetLimitBytes()
+	limitSet := 0 < limitBytes
```
fatih: I think we don't need `requiredSet` and `limitSet` variables as it makes this function complicated. One because it's not a valid value:

```
// The capacity of the storage space in bytes. To specify an exact size,
// `required_bytes` and `limit_bytes` SHALL be set to the same value. At
// least one of the these fields MUST be specified.
type CapacityRange struct {
	// Volume MUST be at least this big. This field is OPTIONAL.
	// A value of 0 is equal to an unspecified field value.
	// The value of this field MUST NOT be negative.
	RequiredBytes int64 `protobuf:"varint,1,opt,name=required_bytes,json=requiredBytes" json:"required_bytes,omitempty"`
	// Volume MUST not be bigger than this. This field is OPTIONAL.
	// A value of 0 is equal to an unspecified field value.
	// The value of this field MUST NOT be negative.
	LimitBytes           int64    `protobuf:"varint,2,opt,name=limit_bytes,json=limitBytes" json:"limit_bytes,omitempty"`
```

It can't be negative, if it's negative, we should return an error immediately. Once we return an error if any of them are negative:

```go
	if 0 < requiredBytes || 0 < limitBytes {
		return 0, fmt.Errorf("limit %v or required %v bytes can't be negative", limitBytes, requiredBytes)
	}
```

we can then remove all `limitSet` and `requiredSet` variables and it would be simpler.

joonas: Unfortunately that's not what these variables are for, they are meant to be used as a way to check whether `RequiredBytes` or `LimitBytes` was provided. Since both of these values can be optional, per the CSI spec, we can't be sure that one, both or either of them set, which is why I thought storing them in a variable that you can simply use to check whether one or the other was set would be convenient.

As it stands today, Kubernetes (via external-provisioner) [only ever sets RequiredBytes](https://github.com/kubernetes-csi/external-provisioner/blob/dcac35d3944fcd9d4f34b0565c3f3f63693c36d5/pkg/controller/controller.go#L419-L421), but this was intended as a way to future proof things should this change.

I'm happy to drop these values if you feel that's better, but as it stands the code wasn't accounting for a whole plethora of cases, which is why I felt that it would be useful to add these.

WDYT?

joonas:
I made some additional error message formatting improvements, the errors you'll now see from trying to provision either a too small or too large volume will be as follows:
```
Events:
  Type       Reason                Age              From                                                                                Message
  ----       ------                ----             ----                                                                                -------
  Normal     ExternalProvisioning  8s (x3 over 8s)  persistentvolume-controller                                                         waiting for a volume to be created, either by external provisioner "dobs.csi.digitalocean.com" or manually created by system administrator
  Normal     Provisioning          8s (x2 over 8s)  dobs.csi.digitalocean.com_csi-do-controller-0_285c902f-ead0-11e8-91b9-02420af43e03  External provisioner is provisioning volume for claim "default/csi-pod-pvc"
  Warning    ProvisioningFailed    8s (x2 over 8s)  dobs.csi.digitalocean.com_csi-do-controller-0_285c902f-ead0-11e8-91b9-02420af43e03  failed to provision volume with StorageClass "do-block-storage": rpc error: code = OutOfRange desc = invalid capacity range: required (953.7Mi) can not be less than minimum supported volume size (1Gi)
```

```
Events:
  Type       Reason                Age                From                                                                                Message
  ----       ------                ----               ----                                                                                -------
  Normal     Provisioning          13s (x2 over 13s)  dobs.csi.digitalocean.com_csi-do-controller-0_285c902f-ead0-11e8-91b9-02420af43e03  External provisioner is provisioning volume for claim "default/csi-pod-pvc"
  Warning    ProvisioningFailed    13s (x2 over 13s)  dobs.csi.digitalocean.com_csi-do-controller-0_285c902f-ead0-11e8-91b9-02420af43e03  failed to provision volume with StorageClass "do-block-storage": rpc error: code = OutOfRange desc = invalid capacity range: required (17Ti) can not exceed maximum supported volume size (16Ti)
  Normal     ExternalProvisioning  11s (x4 over 13s)  persistentvolume-controller                                                         waiting for a volume to be created, either by external provisioner "dobs.csi.digitalocean.com" or manually created by system administrator
```

driver/controller.go
fatih: It's ok, thanks for the explanation Joonas. PR looks fine 👍

PR #102, opened 2018-11-12
 csi: abstract client for better tests

fatih:
The previous approach of faking the Server was not scalable and good enough for us, especially now that we're trying to add Snapshots. This is better because we can mock everything.


nanzhong:
lgtm

PR #104, opened 2018-11-19
csi: revisit rbac rules, add Snapshotter sidecar and roles

fatih:


Makefile
```diff
@@ -47,7 +47,7 @@ test:
 test-integration:
 
 	@echo "==> Started integration tests"
-	@env GOCACHE=off go test -v -tags integration ./test/...
```
aybabtme: why change this?

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -118,6 +118,20 @@ provisioner: dobs.csi.digitalocean.com
 
 ---
 
+# NOTE(arslan): this will probably fail , because the CRD is created via the
+# csi-snapshotter sidecar, that is part of the csi-do-controller statefulset.
+# We need to create this seperately.
```
aybabtme: Add to this comment some documentation about how other devs need to proceed to get this working (eg "apply this section separately after XYZ")

Makefile
fatih: Because go modules won't work if you turn of GOCACHE

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
fatih: This comment will be removed in upcoming PR's it's just here to remind myself.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -190,46 +216,140 @@ spec:
...
+    resources: ["volumesnapshotclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
```
aybabtme: why do you need these permissions to create snapshots?

fatih: The `csi-snapshotter` sidecar needs them to create the `VolumeSnapshot` and `VolumeSnapshotClass` custom resource definitions. 

These are pulled from the appropriate repo, each sidecar now contains the RBAC rules it needs to operate, as an example for the above: https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/rbac.yaml

aybabtme:
:shipit:

PR #105, opened 2018-11-20
csi: add snapshot tests

fatih:
Please checkout the second commit. The first commit is vendoring the client-go and csi-snapshotter packages, which are needed for adding the tests. 

The test is simply just creating a snapshot from an existing pvc and then later restores it back and deploys another pod with restored volume.


test/kubernetes/integration_test.go
```diff
@@ -345,6 +349,176 @@ func TestPod_Multi_Volume(t *testing.T) {
...
+		},
+	}
+
+	t.Log("Creating pvc")
+	_, err = client.CoreV1().PersistentVolumeClaims(namespace).Create(pvc)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("Waiting pod %q to be running ...\n", pod.Name)
```
nanzhong:
```suggestion
	t.Logf("Waiting for pod %q to be running ...\n", pod.Name)
```

PR #113, opened 2018-12-03
csi: v1.0.0

fatih:
CSI v1.0.0 is released (\o/): https://github.com/container-storage-interface/spec/releases/tag/v1.0.0.

This is only compatible with Kubernetes v1.13+. With that, there were
many changes and updates so this PR combines all these changes:

* Updated all sidecars to support the new spec
* Update plugin registry dir from `/var/lib/kubelet/plugins/` to `/var/lib/kubelet/plugins_registry/`
* Updated the dependencies to comply with the latest CSI version, such as `github.com/golang/protobuf` to `v1.2.0`


fatih:
Snapshots integration tests are not working for some reasons, looking into it.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -435,7 +434,7 @@ spec:
       hostNetwork: true
       containers:
         - name: driver-registrar
-          image: quay.io/k8scsi/driver-registrar:v0.4.1
+          image: quay.io/k8scsi/driver-registrar:v1.0-canary
```
joonas: Should we use the new [node-driver-registrar](https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v1.0.1) here?

nanzhong: Agree that it seems like we should use `node-driver-registrar` since `driver-registrar` is deprecated.

fatih: Yeap, as I said this was splitted and needs to be changed. However, I"m waiting for the documentation on how to implement and add both the `cluster-` and `node-` driver registrars. The relevant PR that adds the instructions will be added today or tomorrow to the CSI documentation repo.

PR #115, opened 2018-12-17
Add priorityClassName to controller and node plugin

joonas:
This should help prevent the CSI components from being evicted in favor of user workloads, unless they are explicitly given higher priority.

I didn't include these changes to the `0.1.x` releases since those are likely associated with 1.10.x clusters where this may or may not be enabled.


fatih:
Thanks @joonas

PR #117, opened 2018-12-26
Tweak Kubernetes Compatibility table to be more readable

joonas:
Changes the formatting of the `Kubernetes Compatibility` table ever so slightly to make it more readable, like so:
new-table-formatting


fatih:
👍 Thanks Joonas

PR #127, opened 2019-02-13
Remove debug code that causes warnings on nodes

bouk:
We're currently getting the following error on all nodes when the plugin boots up:

```
time="2019-01-14T20:13:52Z" level=warning msg="CSI plugin will not function correctly, please resolve volume limit" error="rpc error: code = Internal desc = couldn't get account information to check volume limit: GET https://api.digitalocean.com/v2/account: 401 Unable to authenticate you."
```

This is because the worker nodes don't have a DigitalOcean access token (because they don't need one) and thus can't access the API. This seems to be some leftover debugging code, so it should be safe to remove.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -159,12 +159,6 @@ func (d *Driver) Run() error {
 		return resp, err
 	}
 
-	// warn the user, it'll not propagate to the user but at least we see if
-	// something is wrong in the logs
-	if err := d.checkLimit(context.Background()); err != nil {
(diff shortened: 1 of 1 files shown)


bouk:
Got some feedback from Fatih: it would be good to leave this in if a token is passed in, but we should skip it if the plugin is started without a token.

fatih:
Closing this as we still want to check the limits on the controller side. Implemented the suggestion @bouk wrote in his second comment: https://github.com/digitalocean/csi-digitalocean/pull/142

PR #129, opened 2019-03-01
Fix support for volume snapshots by setting snapshot id on volume creation.

justinbarrick:
This bug affects 1.0, as well, however, my cluster is on Kubernetes v1.12.2, so I've targeted 0.4.0.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -161,8 +161,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if contentSource != nil {
 		snapshot := contentSource.GetSnapshot()
 		if snapshot != nil {
-			ll.Info("using snapshot as volume source")
-			volumeReq.SnapshotID = snapshot.GetId()
+			ll.WithField("snapshot_id", snapshot.GetSnapshotId()).Info("using snapshot as volume source")
+			volumeReq.SnapshotID = snapshot.GetSnapshotId()
 		}
 	}
 
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -157,6 +157,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		volumeReq.Tags = append(volumeReq.Tags, d.doTag)
 	}
 
+	contentSource := req.GetVolumeContentSource()
+	if contentSource != nil {
+		snapshot := contentSource.GetSnapshot()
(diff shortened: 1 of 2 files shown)


driver/controller.go
```diff
@@ -148,6 +148,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		SizeGigaBytes: size / GB,
 	}
 
+	contentSource := req.GetVolumeContentSource()
+	if contentSource != nil {
+		snapshot := contentSource.GetSnapshot()
+		if snapshot != nil {
+			ll.Info("using snapshot as volume source")
```
fatih: Could you add the snapshot ID as a K/V to the log? As an example:

```
ll.WithField("snapshot_id", snapshot.GetId()).Info("using snapshot as volume source")
```

test/kubernetes/integration_test.go
```diff
@@ -358,6 +358,22 @@ func TestSnapshot_Create(t *testing.T) {
...
+					VolumeMounts: []v1.VolumeMount{
+						{
+							MountPath: "/data",
+							Name:      volumeName,
+						},
+					},
+					Command: []string{
+						"sh", "-c",
+						"echo testcanary > /data/canary",
+					},
```
fatih: Why do we have these init containers? I couldn't figure out why we need them.

fatih:
Thanks @justinbarrick this looks good. Thanks for the fix. I added couple of fixes. Also please open the PR against **master** and not against the release branches. 

Master is our dev branch and we update the release branches from the master.

test/kubernetes/integration_test.go
justinbarrick: Added a comment, but basically, waitForPod will consider it successful once the pod enters the Running state. Regular containers enter the running state once the healthcheck passes, so by using an initcontainer, the pod will never enter the running state if the assertion fails.

justinbarrick:
Addressed comments and based on master.

justinbarrick:
@fatih rebased. is there anything else you need on this?

justinbarrick:
@fatih I’d love to get this one merged. Anything I can do to help it over the finish line?

As it stands, restoring from a volume with digital ocean csi is 100% broken forcing me to operate a fork.

fatih:
Sorry @justinbarrick  I was derailed from my other work related stuff and hadn't chance to look again. Thanks for your patience and contribution.

PR #130, opened 2019-03-13
Tag volumes on Create/Attach

jcodybaker:
Tag DigitalOcean volumes on Create and Attach when the `--do-tag` flag is specified.  This work is being done to improve billing readability for DO managed Kubernetes cluster.


driver/controller.go
```diff
@@ -253,6 +261,8 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 		return nil, err
 	}
 
+	d.tagVolume(ctx, vol)
```
fatih: We're not checking for the error here.

driver/controller.go
```diff
@@ -963,3 +973,46 @@ func validateCapabilities(caps []*csi.VolumeCapability) bool {
...
+		Resources: []godo.Resource{
+			godo.Resource{
+				ID:   vol.ID,
+				Type: godo.VolumeResourceType,
+			},
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, doAPITimeout)
+	resp, err := d.tags.TagResources(ctx, d.doVolTag, tagReq)
```
fatih: I would return immediately here instead of returning below. It makes the code easier to read.

driver/controller.go
```diff
@@ -963,3 +973,46 @@ func validateCapabilities(caps []*csi.VolumeCapability) bool {
...
+	cancel()
```
fatih: I believe you should call this in a `defer`, otherwise the context will be cancelled immediately, which is also passed down below.

driver/controller.go
```diff
@@ -963,3 +973,46 @@ func validateCapabilities(caps []*csi.VolumeCapability) bool {
...
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
+		// If the tag doesn't exist, create it and retry.
+		err = d.createTag(ctx)
+		if err != nil {
+			return err
+		}
+		ctx, cancel = context.WithTimeout(ctx, doAPITimeout)
+		resp, err = d.tags.TagResources(ctx, d.doVolTag, tagReq)
+		cancel()
+	}
```
fatih: I couldn't understand what this code is meant to do. If we create a tag, why do we check for it's existence again? 
Also seems like we're not checking `resp` here and just continue. Finally, we should return from the error check immediately and not continue. I know you're overriding the `err` here, but I feel like that makes the code open for future bugs.

driver/controller.go
```diff
@@ -963,3 +973,46 @@ func validateCapabilities(caps []*csi.VolumeCapability) bool {
...
+}
+
+func (d *Driver) createTag(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, doAPITimeout)
+	defer cancel()
+	_, _, err := d.tags.Create(ctx, &godo.TagCreateRequest{
+		Name: d.doVolTag,
+	})
+	return err
+}
```
fatih: Is there a reason we're wrapping this call here? We're using it in only one place and I don't think we should create a function for this.

driver/driver.go
```diff
@@ -106,6 +112,7 @@ func NewDriver(ep, token, url string) (*Driver, error) {
 	})
 
 	return &Driver{
+		doVolTag: os.Getenv(doVolTagEnv),
```
fatih: We shouldn't assign it here. Instead this should be passed to `cmd/do-csi-plugin` via a flag, such as `--do-tag`. And then in our controller we would use the still the same environment variable in this form:

```
        - name: csi-do-plugin
          image: digitalocean/do-csi-plugin:v1.0.0
          args :
            - "--endpoint=$(CSI_ENDPOINT)"
            - "--token=$(DIGITALOCEAN_ACCESS_TOKEN)"
            - "--url=$(DIGITALOCEAN_API_URL)"
            - "--do-tag=$DIGITALOCEAN_TAG_VOLUMES)"
```

This is to prevent any side effects on the binary and `cmd/do-csi-plugin` should the only place we pass in external values)

fatih:
Also I think vendoring went wrong and our build is also failing. I would suggest to first upgrade `godo` in a separate PR, pass the CI build and then open a new PR with the tag changes. We shouldn't mix dependency additions and feature improvements in a single PR.

driver/controller.go
jcodybaker: Addressed.  In an earlier iteration I was contemplating making these updates asynchronously so as to not scuttle a volume attach.  I changed that approach, but forgot to handle the error value up here.  

I wasn't sure what the best code would be, so I went with `internal`. https://github.com/digitalocean/csi-digitalocean/pull/130/commits/eea3ef3c35041711cf61ed12ee2da965322e7ae8#diff-aa5513ce5e9cac724e0d7994151fed94R267

driver/controller.go
jcodybaker: When using the TagsService directly (as opposed to as an argument to `CreateVolume`) the tag must exist before you can assign resources (like a volume) to it.  We get a 404 if the tag doesn't exist.  In that case we create the tag, and then retry the `TagResources` which applies the tag to our volume.

driver/controller.go
jcodybaker: I originally did this to try to cleanup some of the multiple context mess in `tagVolume`.  You're right to suggest pulling this into that function.  Done.

driver/driver.go
jcodybaker: Done

driver/controller.go
jcodybaker: Good catch on the context reuse.  I was meaning to use the parent context for each API call rather than chaining them like that.  Tests would have missed that since we faked the TagService.

I struggled on the right way to deal with context here.  From my perspective, as long as we're making progress we should reset the timeout for each call.  So I didn't want to use a single context/cancel. 

I had originally opted to leave the `cancel()` bare rather than in a defer because I thought deferring a cancel() and then overriding that cancel function-ref with a new cancel for the next call was confusing.  `defer cancel()` dereferences the function-ref at the line where the defer is; not the end of the function.  Using the defer does mean those first two timeouts might fire (with no effect) after their associated API call completes but before the defer cancels them.  The only harm is a fairly trivial amount of inefficiency in exchange for a lot of clarity.

driver/controller.go
```diff
@@ -963,3 +977,46 @@ func validateCapabilities(caps []*csi.VolumeCapability) bool {
...
+	})
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel = context.WithTimeout(parentCtx, doAPITimeout)
+	defer cancel()
+	_, err = d.tags.TagResources(ctx, d.doTag, tagReq)
+	return err
+}
```
fatih: Seems like instead of setting individual timeouts for each request, we can set the `client` timeout for our default `http.Client`. This way you don't have to set an explicit timeout for each request. To do that, we need to pass a custom http.Client to the oauth2 library:

```go
// the oauth2 pkg only allows us to modify the http.Client via a ctx key
httpClient := &http.Client{Timeout: doAPITimeout}
ctx = context.WithValue(ctx, oauth2.HTTPClient, httpClient)

resp, err := d.tags.TagResources(ctx, d.doTag, tagReq)
if resp == nil || resp.StatusCode != http.StatusNotFound {
	// either success or irrecoverable failure
	return err
}

// godo.TagsService returns 404 if the tag has not yet been
// created, if that happens we need to create the tag
// and then retry tagging the volume resource.
_, _, err = d.tags.Create(parentCtx, &godo.TagCreateRequest{
	Name: d.doTag,
})
if err != nil {
	return err
}

_, err = d.tags.TagResources(ctx, d.doTag, tagReq)
return err
```

fatih:
Please merge the vendor changes and rebase on top of it before merging this 👍

driver/controller.go
jcodybaker: @fatih and I discussed this.  The above approach may not viable upon closer inspection.  https://github.com/golang/oauth2/blob/master/oauth2.go#L348 seems to show that the http.Client is only extracted from the context in `oauth2.NewClient`.

PR #142, opened 2019-04-19
driver: don't check limits on worker nodes

fatih:
We're checking the volume limit during the driver startup. However
because the driver is used both for the controller and node part, it
fails during the node startup because we're deploying the driver to the
nodes without a token. The DO token is only needed by the controller.

The `checkLimit()` function would therefore print when deployed to the
nodes:

```
time="2019-01-14T20:13:52Z" level=warning msg="CSI plugin will not
function correctly, please resolve volume limit" error="rpc error: code
= Internal desc = couldn't get account information to check volume
limit: GET https://api.digitalocean.com/v2/account: 401 Unable to
authenticate you."
```

This PR fixes that to not to call `checkLimit` if the token is not
passed, which is a valid case for drivers deployed to the worker nodes.

supersedes #127

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -58,6 +58,7 @@ type Driver struct {
 	nodeId   string
 	region   string
 	doTag    string
+	hasToken bool
 
 	srv     *grpc.Server
 	log     *logrus.Entry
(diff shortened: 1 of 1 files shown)


driver/driver.go
```diff
@@ -58,6 +58,7 @@ type Driver struct {
 	nodeId   string
 	region   string
 	doTag    string
+	hasToken bool
```
timoreimann: How about calling this `isController`? I think that would make the code a bit easier to understand and reduce the degree of comments we need below.

fatih: Make sense, pushed some updates. I'll squash afterwards 👍

PR #144, opened 2019-04-22
CSI spec v1.1.0

fatih:
This PR updates the CSI spec to `v1.1.0` and the sidecars to their respective largest versions.

Vendoring the code separately is a breaking change, hence it had to be here. Each commit is self contained, to review, read them one by one. 

* Update the CSI version to `v1.1.0` and the csi-test library to `v2.0.0`
* Satisfy the new controller interface
* Fix/improve the code according to the new additions in the csi-test

- [x] Update sidecars to their respective largest versions compatible with CSI Spec `v1.1.0`
- [x] Create a custom `CSIDriver` because `cluster-driver-registrar` is deprecated and no longer will create it for us (https://github.com/kubernetes/kubernetes/pull/74283)
- [x] Test the changes for K8S `v1.14`


driver/controller.go
```diff
@@ -158,12 +159,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	contentSource := req.GetVolumeContentSource()
-	if contentSource != nil {
-		snapshot := contentSource.GetSnapshot()
-		if snapshot != nil {
-			ll.WithField("snapshot_id", snapshot.GetSnapshotId()).Info("using snapshot as volume source")
-			volumeReq.SnapshotID = snapshot.GetSnapshotId()
+	if contentSource != nil && contentSource.GetSnapshot() != nil {
+		snapshotID := contentSource.GetSnapshot().GetSnapshotId()
```
adamwg: Should we validate that the snapshot ID is not empty? Not sure whether the API call below will return a `NotFound` or some other status when passed an empty ID.

fatih: Good poing @adamwg. I think it'll be already validated but I'll make sure to test that as well.

driver/controller.go
```diff
@@ -790,6 +810,10 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	return listResp, nil
 }
 
+func (d *Driver) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
+	return nil, errors.New("not implemented yet")
```
jcodybaker: Should this return `codes.Unimplemented` rather than a generic go error?

driver/controller.go
fatih: Fixed with: https://github.com/digitalocean/csi-digitalocean/pull/144/commits/46155ff61267529f745e7e41b9f9e838167f87f1

driver/driver_test.go
```diff
@@ -97,21 +96,9 @@ func TestDriverSuite(t *testing.T) {
...
-	mntStageDir, err := ioutil.TempDir("", "mnt-stage")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.RemoveAll(mntStageDir)
-
 	cfg := &sanity.Config{
-		StagingPath: mntStageDir,
-		TargetPath:  mntDir,
+		TargetPath:  os.TempDir() + "/csi-target",
```
jcodybaker: For the benefit of others who are curious about this change.  https://github.com/kubernetes-csi/csi-test/blob/e1f2421c7f0a0cc7f9942fea1192392f253891db/pkg/sanity/sanity.go#L56-L58

PR #148, opened 2019-04-22
Release version 1.0.1

jcodybaker:


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## unreleased
 
+## v1.1.0 - 2019.04.22
```
fatih: nit: should be updated now that we're at 23

fatih:
If we release this as `v1.1.0` we have to release the CSI `v1.1.0` changes as `v1.2.0`. I usually try to match them, but then I'm not sure how meaningful it would be to release this as `v1.0.1`.  I think this is ok.Once we create this, we should create a new `release-1.0.0` branch from the tip of master and lock it against force pushes (just like we do for others). After that master will be dedicated to the CSI spec `v1.1.0` (K8s `v1.14.0`)

fatih:
lgtm

Issue #155, opened 2019-05-05
CSI plugin fails to authenticate after upgrade of kubernetes from 13 (csi 0.4.0) to 14 (csi 1.1.0) 


krasaee:
### What did you do? (required. The issue will be **closed** when not provided.)

1. upgraded kubernetes nodes from version 13 to 14 using kubeadm
2. upgraded csi plugin from 0.4.0 by simply applying v1.1.0 via kubectl
3. everything appeared to work for a while but now receiving (401 Unable to authenticate you)
4. manually removed traces of do-csi, daemonsets, roles and so forth. 
5. deleted secret and applied new secret, checked using doctl that it is correct after running "export DIGITALOCEAN_ACCESS_TOKEN=$(kubectl get secrets -n kube-system digitalocean -o json | jq -r '.data["access-token"]'| base64 --decode)"
and then doctl compute volume list
7. created simple pvc but receiving 401 unauthorized
8. triple checked to do secret

### What did you expect to happen?

1. expect pvc to provision as it was working with 0.4.0 of csi plugin on kube 13

### Configuration (**MUST** fill this out):

* system logs:

`kubectl logs csi-do-controller-0 -n kube-system csi-provisioner`

```
I0505 10:16:40.392423       1 controller.go:1196] provision "default/csi-pvc" class "do-block-storage": started
I0505 10:16:40.395540       1 controller.go:442] CreateVolumeRequest {Name:pvc-baef1cc8-6f1e-11e9-9f09-1e6b332d2b41 CapacityRange:required_bytes:5368709120  VolumeCapabilities:[mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > ] Parameters:map[] Secrets:map[] VolumeContentSource:<nil> AccessibilityRequirements:<nil> XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
I0505 10:16:40.395622       1 connection.go:180] GRPC call: /csi.v1.Controller/CreateVolume
I0505 10:16:40.397285       1 event.go:209] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"csi-pvc", UID:"baef1cc8-6f1e-11e9-9f09-1e6b332d2b41", APIVersion:"v1", ResourceVersion:"37126277", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "default/csi-pvc"
[...]

PR #159, opened 2019-05-17
Add missing access token references to v1.1.0 Kubernetes release

### LINKED ISSUE #158: Resources manifest for 1.1.0 is broken and results in 401
Hello. Sorry for not following this project's issue reporting guidelines, I just wanted to point to the nasty quirk with resources manifest for the current release - https://github.com/digitalocean/csi-digitalocean/blob/master/deploy/kubernetes/releases/csi-digitalocean-v1.1.0.yaml - `csi-do-plugin` container spec declaration for both `csi-do-controller` and `csi-do-node` doesn't define arguments and secrets correctly:

- `--token` argument is missing for  `csi-do-node`'s `csi-do-plugin`
- `csi-do-controller`'s `csi-do-plugin` doesn't include secret declaration for `DIGITALOCEAN_ACCESS_TOKEN`

Absence of these two results in 401 responses from DO API. I've fixed this in my environment by adding needed declarations as per manifests for previous releases.

forestlovewood:
This is a straightforward change that adds the access token references missing from the v1.1.10 Kubernetes release definition. Resolves #158 and probably #155.

### DIFF
--- a/deploy/kubernetes/releases/csi-digitalocean-v1.1.0.yaml
+++ b/deploy/kubernetes/releases/csi-digitalocean-v1.1.0.yaml
@@ -179,6 +179,11 @@ spec:
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
             - name: DIGITALOCEAN_API_URL
               value: https://api.digitalocean.com/
+            - name: DIGITALOCEAN_ACCESS_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: digitalocean
+                  key: access-token
           imagePullPolicy: "Always"
           volumeMounts:
             - name: socket-dir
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/releases/csi-digitalocean-v1.1.0.yaml
```diff
@@ -381,6 +386,7 @@ spec:
           image: digitalocean/do-csi-plugin:v1.1.0
           args :
             - "--endpoint=$(CSI_ENDPOINT)"
+            - "--token=$(DIGITALOCEAN_ACCESS_TOKEN)"
```
fatih: This is not needed anymore. The node plugin driver doesn't use the token anymore.

fatih:
Thanks @forestlovewood for fixing it!  We don't pass the token anymore in our managed offering, hence it was also removed from here (by accident). 

We don't need to pass the token to the `node` driver though, if you can remove that line, this is good to merge :+1:

deploy/kubernetes/releases/csi-digitalocean-v1.1.0.yaml
forestlovewood: @fatih I've removed that line from the pull request.

PR #161, opened 2019-06-17
Support XFS

snormore:
Include `xfsprogs` in the base Docker image so that XFS can be used.

Also validated in https://github.com/snormore/doks-examples/tree/master/xfs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## unreleased
 
 * Set a custom user agent for the godo client. [[GH-156](https://github.com/digitalocean/csi-digitalocean/pull/156)]
+* Include `xfsprogs` in the base Docker image so that XFS can be used
 
 ## v1.1.0 - 2019.04.29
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,7 +34,7 @@ bump-version:
 .PHONY: compile
 compile:
 	@echo "==> Building the project"
-	@env CGO_ENABLED=0 GOOS=${OS} GOARCH=amd64 go build -o cmd/do-csi-plugin/${NAME} -ldflags "$(LDFLAGS)" ${PKG} 
+	@docker run --rm -it -e GOOS=${OS} -e GOARCH=amd64 -v ${PWD}/:/app -w /app golang:1.12-alpine sh -c 'apk add git && go build -o cmd/do-csi-plugin/${NAME} -ldflags "$(LDFLAGS)" ${PKG}'
```
fatih: I like this, but I think this should be in a separate PR.

snormore: Good call 👉 https://github.com/digitalocean/csi-digitalocean/pull/162

fatih:
LGTM

PR #163, opened 2019-07-02
Bump version to v1.1.1

waynr:
...also remove unnecessary `rm` call at end of `bump-version` Make target.


waynr:
@jcodybaker done, thanks for catching that!

README.md
```diff
@@ -58,7 +58,7 @@ bumped following the rules below:
       <td>no</td>
     </tr>
     <tr>
-      <td>v1.1.0 - v1.1.x</td>
+      <td>v1.1.1 - v1.1.x</td>
```
jcodybaker: I know the script probably did this, but this should stay 1.1.0, because it's the lower bound version for compatible with 1.14.x.

PR #179, opened 2019-09-13
Allow a custom driver name to be specified on the commandline

adamwg:
The name of our CSI driver changed between the 0.2 and 0.4 releases, from `com.digitalocean.csi.dobs` to `dobs.csi.digitalocean.com`. This means that when upgrading a cluster from Kubernetes 1.11 (which uses CSI plugin 0.2) to Kubernetes 1.12 (which uses CSI plugin 0.4) volumes created before the upgrade have the old driver name in various annotations and spec fields. This prevents the new driver from managing the volume, leading to broken workloads.

To handle the upgrade case, allow a custom driver name to be specified on the commandline. This way, a cluster that was once running 1.11 and thus uses the old name can continue using the old name. By default the new name will be used.

### DIFF
--- a/cmd/do-csi-plugin/main.go
+++ b/cmd/do-csi-plugin/main.go
@@ -27,11 +27,12 @@ import (
 
 func main() {
 	var (
-		endpoint = flag.String("endpoint", "unix:///var/lib/kubelet/plugins/"+driver.DriverName+"/csi.sock", "CSI endpoint")
-		token    = flag.String("token", "", "DigitalOcean access token")
-		url      = flag.String("url", "https://api.digitalocean.com/", "DigitalOcean API URL")
-		doTag    = flag.String("do-tag", "", "Tag DigitalOcean volumes on Create/Attach")
-		version  = flag.Bool("version", false, "Print the version and exit.")
+		endpoint   = flag.String("endpoint", "unix:///var/lib/kubelet/plugins/"+driver.DefaultDriverName+"/csi.sock", "CSI endpoint")
+		token      = flag.String("token", "", "DigitalOcean access token")
+		url        = flag.String("url", "https://api.digitalocean.com/", "DigitalOcean API URL")
+		doTag      = flag.String("do-tag", "", "Tag DigitalOcean volumes on Create/Attach")
+		driverName = flag.String("driver-name", driver.DefaultDriverName, "Name for the driver")
+		version    = flag.Bool("version", false, "Print the version and exit.")
 	)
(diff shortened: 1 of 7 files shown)


cmd/do-csi-plugin/main.go
```diff
@@ -27,11 +27,12 @@ import (
 
 func main() {
 	var (
-		endpoint = flag.String("endpoint", "unix:///var/lib/kubelet/plugins/"+driver.DriverName+"/csi.sock", "CSI endpoint")
-		token    = flag.String("token", "", "DigitalOcean access token")
-		url      = flag.String("url", "https://api.digitalocean.com/", "DigitalOcean API URL")
-		doTag    = flag.String("do-tag", "", "Tag DigitalOcean volumes on Create/Attach")
-		version  = flag.Bool("version", false, "Print the version and exit.")
+		endpoint   = flag.String("endpoint", "unix:///var/lib/kubelet/plugins/"+driver.DefaultDriverName+"/csi.sock", "CSI endpoint")
```
adamwg: Note: We're using the default driver name here, not the commandline-provided one. The endpoint socket path just needs to match what's in the deployment manifest, not the driver name.

driver/node.go
```diff
@@ -95,7 +90,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		"method":              "node_stage_volume",
 	})
 
-	_, ok := req.VolumeContext[annNoFormatVolume]
+	_, ok := req.VolumeContext[d.annNoFormatVolume]
```
jcodybaker: I wonder if should allow both old and new annotations here.  That might make user-facing documentation easier and reduce the pain caused by confusing the two.

adamwg: Good idea. I've implemented this in the latest push.

I hard-coded the new and legacy driver names for this rather than using the flag - seems sensible given that we're supporting both, but let me know if you have a nicer way to do it.

driver/node.go
```diff
@@ -40,11 +40,18 @@ const (
...
 	// See: https://www.digitalocean.com/docs/volumes/overview/#limits
 	maxVolumesPerNode = 7
+)
 
+var (
 	// This annotation is added to a PV to indicate that the volume should be
 	// not formatted. Useful for cases if the user wants to reuse an existing
-	// volume.
-	annNoFormatVolume = DriverName + "/noformat"
+	// volume. We support using either the legacy driver name
```
snormore: :+1:

PR #188, opened 2019-10-08
Make it easy to test CSI driver changes using a DOKS cluster

adamwg:
Currently there's no easy way to test a change to this CSI driver using a DOKS cluster, since DOKS clusters run the CSI controller as a managed component that cannot be modified by users.

This PR introduces a process that can be used to easily test a development version of the CSI driver on a DOKS cluster. We use `kustomize` to modify the provided dev deployment manifest, renaming the storage class, provisioner, and workloads so that the dev version can run alongside the pre-installed version. The integration tests are then run using the dev storage class rather than the production one.

In testing this, I found a problem with the dev manifest (the `DIGITALOCEAN_ACCESS_TOKEN` secret was being provided to the wrong container), so that's fixed here as well. I also noticed that `make test-integration` was using cached results after the first run, which doesn't seem desirable for integration tests; I've used `-count 1` to avoid that.

### DIFF
--- a/test/kubernetes/deploy/README.md
+++ b/test/kubernetes/deploy/README.md
@@ -1,0 +1,54 @@
+# Deploying and Testing a Development Version
+
+The most convenient way to test changes to csi-digitalocean is on a
+[DigitalOcean Kubernetes (DOKS)](https://www.digitalocean.com/products/kubernetes/)
+cluster. However, since DOKS clusters run the DigitalOcean CSI controller as a
+managed component that cannot be modified by the user, deploying with the
+example manifests will not work.
+
+To test a development version on a DOKS cluster, do the following:
+
+1. Create a DOKS cluster on the latest version:
+   ```console
+   $ doctl k8s cluster create csi-integration-test
+   Notice: cluster is provisioning, waiting for cluster to be running
+   ..
+   ```
+   Wait for it to finish creating; your kubeconfig will automatically be updated.
+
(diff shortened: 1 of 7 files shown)


test/kubernetes/deploy/deploy.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if ! command -v kustomize &> /dev/null; then
```
timoreimann: `&>` is Bash v4 only. Better to use `2>&1`. (See also [this SO answer](https://stackoverflow.com/a/876267/420061).)

Similar below.

test/kubernetes/deploy/README.md
```diff
@@ -0,0 +1,53 @@
+# Deploying and Testing a Development Version
+
+The most convenient way to test changes to csi-digitalocean is on a DOKS
```
snormore: Link to the product page for the first instance of "[DOKS](https://www.digitalocean.com/products/kubernetes/)"?

test/kubernetes/deploy/deploy.sh
adamwg: Aha, good catch. Fixed.

test/kubernetes/deploy/README.md
adamwg: Good idea - expanded DOKS to DigitalOcean Kubernetes in the first instance, and also added the link.

PR #193, opened 2019-10-16
Implement ControllerExpandVolume and NodeExpandVolume to resize volumes automatically

varshavaradarajan:


driver/node.go
```diff
@@ -337,7 +348,21 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	volumeID := req.GetVolumeId()
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
+	}
+
+	r := resizefs.NewResizeFs(&mount.SafeFormatAndMount{
+		Interface: mount.New(""),
+		Exec:      mount.NewOsExec(),
+	})
+	// todo: find device path
```
varshavaradarajan: @adamwg - I wasn't sure how to find the device path needed to resize: https://godoc.org/k8s.io/kubernetes/pkg/util/resizefs
Can you please help me out?

adamwg: Doesn't look like we get the volume name or device path in the request, so we'll probably have to infer it from the mount path. Looks like [`mount.GetDeviceNameFromMount`](https://godoc.org/k8s.io/kubernetes/pkg/util/mount#GetDeviceNameFromMount) might be helpful here.

deploy/kubernetes/releases/csi-digitalocean-v1.1.3.yaml
```diff
@@ -0,0 +1,510 @@
...
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
+              value: /var/lib/csi/sockets/pluginproxy/csi.sock
+          imagePullPolicy: Always
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: csi-resizer
+          image: quay.io/k8scsi/csi-resizer:canary
```
adamwg: I think we should use a numbered release here rather than `canary`. Looks like the latest is currently `v0.3.0`.

driver/controller.go
```diff
@@ -806,11 +806,30 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	return listResp, nil
 }
 
+// ControllerExpandVolume is called from the resizer to increase the volume size.
 func (d *Driver) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
-	d.log.WithField("method", "controller_expand_volume").
-		Info("controller expand volume called")
```
adamwg: We should retain this log and add more information to it, similar to what we do in the [controller Create RPC](https://github.com/digitalocean/csi-digitalocean/blob/master/driver/controller.go#L114). These messages are super helpful when debugging CSI issues.

driver/controller.go
```diff
@@ -806,11 +806,30 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	volID := req.GetVolumeId()
 
-	return nil, status.Error(codes.Unimplemented, "")
+	if len(volID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID missing in request")
```
adamwg: Nit: we start error messages lower-case unless the first word is a proper noun or method name. So, `volume ID missing in request`.

(Same comment for a few of the errors in the node driver.)

driver/controller.go
```diff
@@ -806,11 +806,30 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	}
+
+	resizeBytes, err := extractStorage(req.GetCapacityRange())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	action, _, err := d.storageActions.Resize(ctx, req.GetVolumeId(), int(resizeBytes), d.region)
```
adamwg: `Resize` takes its size argument in GiB - I think `resizeBytes / giB` would be correct here.

driver/controller.go
```diff
@@ -806,11 +806,30 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "StoragesActions.Resize returned error: %s", err.Error())
+	}
+	if action != nil {
+		if err := d.waitAction(ctx, req.VolumeId, action.ID); err != nil {
+			return nil, err
+		}
+	}
+
+	return &csi.ControllerExpandVolumeResponse{CapacityBytes: req.GetCapacityRange().GetRequiredBytes(), NodeExpansionRequired: true}, nil
```
adamwg: I think we should use `resizeBytes` as the return here, since `extractStorage` has some slightly complex logic for choosing the size we end up resizing to.

driver/node.go
```diff
@@ -337,7 +348,27 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to get device path for %q: %v", req.GetVolumePath(), err)
+	}
+
+	r := resizefs.NewResizeFs(&mount.SafeFormatAndMount{
+		Interface: mounter,
+		Exec:      mount.NewOsExec(),
+	})
+
+	if _, err := r.Resize(devicePath, req.GetVolumePath()); err != nil {
```
timoreimann: The [CSI spec distinguishes between online and offline resizing](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerexpandvolume). Do I understand correctly that we support the online variant?

Either way, I believe we should expose our capabilities in `GetPluginCapabilities` by what I think should look like this (assuming we support _online_):

```go
Type: &csi.PluginCapability_VolumeExpansion_{
  VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
    Type: csi.PluginCapability_VolumeExpansion_ONLINE,
  },
}
```

driver/controller.go
```diff
@@ -806,11 +806,33 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	volID := req.GetVolumeId()
 
-	return nil, status.Error(codes.Unimplemented, "")
+	if len(volID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume ID missing in request")
+	}
+	ll.Info("extracting storage from requested capacity range")
+	resizeBytes, err := extractStorage(req.GetCapacityRange())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
```
timoreimann: We should probably mimic the pre-existing error handling of the `extractStorage` call and return a `codes.OutOfRange` code in order to [fulfill the CSI spec with regards to the expected error for _"Unsupported `capacity_range`"_](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerexpandvolume-errors).

driver/controller.go
```diff
@@ -806,11 +806,33 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	}
+	resizeGigaBytes := resizeBytes / giB
+	ll.WithField("new_volume_size", resizeGigaBytes).Info("attempting to resize volume")
+	action, _, err := d.storageActions.Resize(ctx, req.GetVolumeId(), int(resizeGigaBytes), d.region)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", req.GetVolumeId(), err.Error())
```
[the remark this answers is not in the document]
timoreimann: Oh no that's fine then -- I wasn't aware the storage API provided that level of detail. 👍

varshavaradarajan:
@timoreimann - I'll squash the commits after you approve the changes. Thanks for reviewing this! The only other change that I made apart from what you asked is: https://github.com/digitalocean/csi-digitalocean/pull/193/files#diff-aa5513ce5e9cac724e0d7994151fed94R593

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -168,6 +168,19 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: csi-resizer
+          image: quay.io/k8scsi/csi-resizer:v0.3.0
+          args:
+            - "--v=5"
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election"
```
timoreimann: I think we only specify a single csi-resizer instance, do we?

[The docs](https://github.com/kubernetes-csi/external-resizer#recommended-optional-arguments) say about this parameter (**emphasis** by me):

>`--leader-election:` Enables leader election. **This is mandatory when there are multiple replicas of the same external-resizer running for one CSI driver.** Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.

If our example doesn't specify multiple instances, then we shouldn't need this, right?

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -168,6 +168,19 @@ spec:
...
+          env:
+            - name: ADDRESS
+              value: /var/lib/csi/sockets/pluginproxy/csi.sock
+          imagePullPolicy: "IfNotPresent"
```
timoreimann: nit: FWIW, other image pull policies defined in our manifests use "Always". Not really sure what's better, but might be good to be consistent (unless we have a good reason to diverge).

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -333,6 +346,35 @@ roleRef:
...
+# Resizer must be able to work with end point in current namespace
+# if (and only if) leadership election is enabled
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  namespace: kube-system
+  name: external-resizer-cfg
+rules:
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
```
timoreimann: I suppose these RBAC may not be required if we choose to drop leader election.

deploy/kubernetes/releases/csi-digitalocean-v1.1.3.yaml
```diff
@@ -0,0 +1,510 @@
+# Copyright 2019 DigitalOcean
```
timoreimann: We [generate / update this file dynamically with our bump-version Makefile target](https://github.com/digitalocean/csi-digitalocean/blob/a31f090a350ed24bce07fbd34ed34ae851625fc6/Makefile#L26-L27), so I assume we should not commit it. The next release would overwrite it.

driver/controller_test.go
```diff
@@ -151,3 +156,84 @@ func (f *fakeTagsDriver) TagResources(ctx context.Context, tag string, req *godo
 func (*fakeTagsDriver) UntagResources(context.Context, string, *godo.UntagResourcesRequest) (*godo.Response, error) {
 	panic("not implemented")
 }
+
+func TestControllerExpandVolume(t *testing.T) {
```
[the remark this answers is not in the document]
timoreimann: I believe the test removal part of my comment hasn't been addressed yet. Looking at the upstream tests I linked to above, the tests _successful resize_ and _volume id missing_ are already implemented upstream (starting [here](https://github.com/kubernetes-csi/csi-test/blob/ec1441c6838578b38bf361ed3224155e30ac4cfb/pkg/sanity/controller.go#L2135) and [here](https://github.com/kubernetes-csi/csi-test/blob/ec1441c6838578b38bf361ed3224155e30ac4cfb/pkg/sanity/controller.go#L2104), respectively), so I feel like we don't need to maintain a copy of our own.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -1,17 +1,17 @@
-# Copyright 2018 DigitalOcean
-# 
+# Copyright 2019 DigitalOcean
```
timoreimann: Github doesn't support commenting on lines outside the context of a change, so I'm noting it down here:

Could you please update the API versions of our StatefulSet and DaemonSet manifests from `apps/v1beta2` to `apps/v1` in this file? The former are deprecated in Kubernetes 1.16 and wouldn't work anymore. (See also #202.)

I have already made a note to also update the non-dev release manifests when we do the next release. I'll take care of that.

timoreimann:
I was just thinking that adding a test to our [integration test suite](https://github.com/digitalocean/csi-digitalocean/blob/a31f090a350ed24bce07fbd34ed34ae851625fc6/test/kubernetes/integration_test.go) may also be beneficial; but that can also be done in a follow-up PR.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
varshavaradarajan: I was looking at the imagePullyPolicy for the existing containers. The images for them have versions as tags. So, I don't understand why the policy is set to `Always`.  When there is a change, we have to bump up the tag anyway. So, why download the image every time?

deploy/kubernetes/releases/csi-digitalocean-v1.1.3.yaml
varshavaradarajan: Since I am creating a new container spec for resizer, which doesn't exist in version 1.1.2, how should we handle this? The generate script just copies the  the current version manifest tothe new version. I'll add a condition to not do that when there is a file existing for the new version?

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
timoreimann: That's a fair point, I think it makes sense to stick to `IfNotPresent` then. 👍 

(Technically, I believe `Always` doesn't pull the entire image but checks which layers are missing / diverging and only pull if that's the case. But still no good reason for us to use `Always` .)

deploy/kubernetes/releases/csi-digitalocean-v1.1.3.yaml
timoreimann: The way I think we can handle this is to add the resizer and apply any other changes after we have invoked `bump-version`. Essentially, the change would go in as we do the release.

Going forward, we could have a templated version of our release manifest that we can update as we add features / make changes (i.e. during development) and only populate during the release with the exact tags that we want. But I think we should create an issue for that and do it in a follow-up PR.

WDYT?

varshavaradarajan: Yeah, sounds good! I'll create the issue today. 
For this PR, I will not commit the 1.1.3 manifest, but during the new release, we should keep track of this PR and add the container spec for resizer.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
varshavaradarajan: I'll raise a new PR to change other containers' imagePullPolicy to IfNotPresent. Is that okay?

deploy/kubernetes/releases/csi-digitalocean-v1.1.3.yaml
timoreimann: 👌 I suppose we will be able to copy the relevant updates from the dev manifest that is going to be committed by this PR.

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
timoreimann: Sounds good to me! 👍

timoreimann:
Ugh, looks like we'll also need to update [our kustomization.yaml file](https://github.com/digitalocean/csi-digitalocean/blob/efca1f1a23e08826478b9a65a4359d61a5b97512/test/kubernetes/deploy/kustomization.yaml).

deploy/kubernetes/releases/csi-digitalocean-dev.yaml
```diff
@@ -117,7 +117,7 @@ provisioner: dobs.csi.digitalocean.com
 ##############################################
 
 kind: StatefulSet
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
```
timoreimann: Just realized that the `apps/v1` also requires a [`selector`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#statefulsetspec-v1-apps) under the spec or otherwise `kubectl apply` is going to fail.

timoreimann:
Just tried resizing a PVC with the DO plugin from this PR running but got this after updating the manifest:

```
error: persistentvolumeclaims "csi-pvc" could not be patched: persistentvolumeclaims "csi-pvc" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
```

Are we still missing something, or am I doing it wrong? I tried to update the storage request resource from 5 GiB to 20 GiB in

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"name":"csi-pvc","namespace":"default"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"5Gi"}},"storageClassName":"do-block-storage-dev"}}
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: dobs.csi.digitalocean.com-dev
  creationTimestamp: "2019-11-11T14:58:59Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: csi-pvc
  namespace: default
  resourceVersion: "5045"
  selfLink: /api/v1/namespaces/default/persistentvolumeclaims/csi-pvc
  uid: 64b7855d-1eb0-4861-9111-ea0baa09b905
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  storageClassName: do-block-storage-dev
  volumeMode: Filesystem
  volumeName: pvc-64b7855d-1eb0-4861-9111-ea0baa09b905
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 5Gi
  phase: Bound
```

timoreimann:
Pushed a few commits to address my latest comments.

Still needs more testing, and I think an integration test would definitely be good to have.

test/kubernetes/integration_test.go
```diff
@@ -246,6 +246,138 @@ func TestDeployment_Single_Volume(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("Waiting pod %q to be running ...\n", pod.Name)
+	if err := waitForPod(client, pod.Name); err != nil {
+		t.Error(err)
+	}
+
+	createdPVC, err := client.CoreV1().PersistentVolumeClaims(namespace).Get(claimName, metav1.GetOptions{})
```
timoreimann: The call to create the PVC above also returns the persisted object, so we can use that instead of fetching it again.

test/kubernetes/integration_test.go
```diff
@@ -246,6 +246,138 @@ func TestDeployment_Single_Volume(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// get pv name
```
timoreimann: IMO we can skip the comment, it's obvious from the code.

test/kubernetes/integration_test.go
```diff
@@ -246,6 +246,138 @@ func TestDeployment_Single_Volume(t *testing.T) {
...
+	pvName := createdPVC.Spec.VolumeName
+	pv, err := client.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if pv.Spec.Capacity["storage"] != resource.MustParse("5Gi") {
+		t.Fatal("initial volume size is not equal to requested volume size")
```
timoreimann: Let's include the values we were comparing (`Spec.Capacity["storage"]` and `"5Gi"`) in the output.

Similar in the assertion below.

test/kubernetes/integration_test.go
```diff
@@ -674,6 +806,85 @@ func waitForPod(client kubernetes.Interface, name string) error {
...
 
+// waitForVolumeCapacityChange waits for the given pod name to be running
+func waitForVolumeCapacityChange(client kubernetes.Interface, name string, resourceList v1.ResourceList) error {
+	var err error
+	stopCh := make(chan struct{})
+
+	go func() {
+		select {
+		case <-time.After(time.Minute * 5):
+			err = errors.New("timing out waiting for pv state")
```
timoreimann: Maybe instead of _pv state_, say _PV capacity change_?

test/kubernetes/integration_test.go
```diff
@@ -674,6 +806,85 @@ func waitForPod(client kubernetes.Interface, name string) error {
 	return err
 }
 
+// waitForVolumeCapacityChange waits for the given pod name to be running
```
timoreimann: Copy-pasta GoDoc

test/kubernetes/integration_test.go
```diff
@@ -674,6 +806,85 @@ func waitForPod(client kubernetes.Interface, name string) error {
...
+					return
+				}
+			},
+		})
+
+	controller.Run(stopCh)
+	return err
+}
+
+// waitForVolumeClaimCapacityChange waits for the given pod name to be running
```
timoreimann: Copy-pasta GoDoc

test/kubernetes/integration_test.go
```diff
@@ -674,6 +806,85 @@ func waitForPod(client kubernetes.Interface, name string) error {
...
+func waitForVolumeClaimCapacityChange(client kubernetes.Interface, name string, resourceList v1.ResourceList) error {
+	var err error
+	stopCh := make(chan struct{})
+
+	go func() {
+		select {
+		case <-time.After(time.Minute * 5):
+			err = errors.New("timing out waiting for pvc state")
```
timoreimann: `s/pvc state/PVC capacity change/`?

test/kubernetes/integration_test.go
```diff
@@ -674,6 +806,85 @@ func waitForPod(client kubernetes.Interface, name string) error {
...
+					return
+				}
+				if volume.Status.Phase == v1.VolumeFailed {
+					err = errors.New("Persistent volume status is Failed")
+					close(stopCh)
+					return
+				}
+
+				if volume.Status.Phase == v1.VolumeBound && volume.Spec.Capacity["storage"] != resourceList["storage"] {
+					close(stopCh)
```
timoreimann: If we also store the updated object here and return it after the controller has stopped, we can save the call to the API to fetch the PV resource again at line 355.

Similar for PVC.

test/kubernetes/integration_test.go
```diff
@@ -246,6 +246,138 @@ func TestDeployment_Single_Volume(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	resizedPVC, err := client.CoreV1().PersistentVolumeClaims(namespace).Get(claimName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if resizedPVC.Status.Capacity["storage"] != resource.MustParse("6Gi") {
+		t.Fatal("file system resize failed: claim status capacity is not equal to requested capacity")
```
timoreimann: consistency nit: the wording / phrasing is slightly different from line 361; but maybe that's desirable?

test/kubernetes/integration_test.go
```diff
@@ -246,6 +246,138 @@ func TestDeployment_Single_Volume(t *testing.T) {
...
+	}
+
+	t.Log("Finished!")
```
timoreimann: `t.Log()` statements only emit messages when a test fails, so we can drop this one.

driver/controller.go
```diff
@@ -806,11 +807,42 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	}
+	ll.Info("extracting storage from requested capacity range")
+	resizeBytes, err := extractStorage(req.GetCapacityRange())
+	if err != nil {
+		return nil, status.Errorf(codes.OutOfRange, "ControllerExpandVolume invalid capacity range: %v", err)
+	}
+	resizeGigaBytes := resizeBytes / giB
+
+	if resizeGigaBytes <= volume.SizeGigaBytes {
+		return &csi.ControllerExpandVolumeResponse{CapacityBytes: volume.SizeGigaBytes * giB, NodeExpansionRequired: true}, nil
```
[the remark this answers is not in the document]
varshavaradarajan: Consider this scenario: 
- user creates a pvc initially with 5Gi capacity. 
- A dynamic PV is created with 5Gi capacity
- Using the control panel, the volume's capacity is increased to 20Gi
- The PVC is then edited to increase the capacity to 10Gi
- NodeExpandVolume should be called to resize the FS to (10Gi | 20Gi), right? 

Increasing the volume size in the control panel does not mean that the PVC capacity will increase. When the user requests that the PVC capacity be increased, we should make a call to `NodeExpandVolume`. We discussed that this shouldn't cause an error while requesting a resize. 

We were discussing whether the `CapacityBytes` we return here should be the what the user requested or volume capacity. I have provided the volume capacity here so that the PV and PVC capacities become consistent when the resize is requested. Not sure if it is alright for the two capacities to be inconsistent for dynamically provisioned PVs.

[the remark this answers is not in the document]
timoreimann: Thanks for explaining, that does make sense. Perhaps we should test this scenario manually to make sure it really works as expected and doesn't break things. WDYT?

(And if it works as expected, we should add a quick comment above the line in question to describe the reasoning to future code readers.)

test/kubernetes/integration_test.go
varshavaradarajan: Until the PVC is bound to the pod, we won't get the volume name in the spec. So, we need to fetch the PVC after the pod is ready.

driver/controller.go
```diff
@@ -806,11 +807,45 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+	resizeBytes, err := extractStorage(req.GetCapacityRange())
+	if err != nil {
+		return nil, status.Errorf(codes.OutOfRange, "ControllerExpandVolume invalid capacity range: %v", err)
+	}
+	resizeGigaBytes := resizeBytes / giB
+
+	if resizeGigaBytes <= volume.SizeGigaBytes {
+		// even if the volume is resized independently from the control panel, we still need to resize the node fs when resize is requested
+		// in this case, the claim capacity will be resized to the volume capacity, requested capcity will be ignored to make the PV and PVC capacities consistent 
+		ll.Info("skip volume resizing: current volume size (%v) exceeds requested volume size (%v Gi). Node FS resize will be performed to resize volume claim.", formatBytes(resizeBytes), volume.SizeGigaBytes)
```
timoreimann: CI is complaining because `logrus` expects parameters to be attached as key-value results.

PR #197, opened 2019-11-02
Implement NodeGetVolumeStats RPC

### LINKED ISSUE #134: Volume stats missing
### What did you do? (required. The issue will be **closed** when not provided.)
Setup kube-prometheus (https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus)

### What did you expect to happen?
Being able to see PV stats, like diskspace free etc. There is an dashboard for this already added, but data was empty. Checked out prometheus data and didn't find any metrics named kubelet_volume_stats_*

Searched around, and it looks like from my understanding that the driver should implement this.

https://stackoverflow.com/questions/44718268/how-to-monitor-disk-usage-of-kubernetes-persistent-volumes?rq=1

* CSI Version:
latest managed DO cluster

* Kubernetes Version:
1.3.14 managed DO cluster

timoreimann:
Exposing node capacity statistics facilitates features like alerting on high disk usage.

I ran my implementation on a cluster with Prometheus on and was able to retrieve metrics with the `kubelet_volume_*` prefix, e.g., `kubelet_volume_stats_capacity_bytes`. So this seems to work. I'll do some more testing for sure though before merging.

Fixes #134

### DIFF
--- a/driver/node.go
+++ b/driver/node.go
@@ -285,23 +285,29 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 
 // NodeGetCapabilities returns the supported capabilities of the node server
 func (d *Driver) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetCapabilitiesRequest) (*csi.NodeGetCapabilitiesResponse, error) {
-	// currently there is a single NodeServer capability according to the spec
-	nscap := &csi.NodeServiceCapability{
-		Type: &csi.NodeServiceCapability_Rpc{
-			Rpc: &csi.NodeServiceCapability_RPC{
-				Type: csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
+	nscaps := []*csi.NodeServiceCapability{
+		&csi.NodeServiceCapability{
+			Type: &csi.NodeServiceCapability_Rpc{
+				Rpc: &csi.NodeServiceCapability_RPC{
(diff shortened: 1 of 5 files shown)


driver/node.go
```diff
@@ -327,10 +333,57 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
 // NodeGetVolumeStats returns the volume capacity statistics available for the
 // the given volume.
 func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	d.log.WithField("method", "node_get_volume_stats").
-		Info("node get volume stats called")
+	ll := d.log.WithField("method", "node_get_volume_stats")
+	ll.Info("node get volume stats called")
 
-	return nil, status.Error(codes.Unimplemented, "")
+	if req.VolumeId == "" {
```
adamwg: Doesn't look like we actually use the `VolumeId`. Are we validating it just because that's what's in the CSI spec?

timoreimann: That's correct: the field is [marked as required per the CSI spec](https://github.com/container-storage-interface/spec/blob/4731db0e0bc53238b93850f43ab05d9355df0fd9/spec.md#nodegetvolumestats), and [the upstream unit tests codify this requirement](https://github.com/digitalocean/csi-digitalocean/blob/17343cf952e13ed59c624b0fceef8bed508212c1/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go#L399).

Issue #206, opened 2019-11-08
Derive release manifests from latest dev versions


timoreimann:
We currently generate a new set of release manifests [by copying the latest release manifests and bumping the CSI plugin tags](https://github.com/digitalocean/csi-digitalocean/blob/a31f090a350ed24bce07fbd34ed34ae851625fc6/Makefile#L26-L27).

This is problematic in two regards:

1. Changes to the release manifests other than tag bumps cannot be made unless we decide to modify already released copies (i.e., the latest instance of `csi-digitalocean-vX.Y.Z.yaml` [here](https://github.com/digitalocean/csi-digitalocean/tree/a31f090a350ed24bce07fbd34ed34ae851625fc6/deploy/kubernetes/releases)). The latter is somewhat concerning because it creates a discrepancy between the release manifests in master and on the tagged commit, which is a potential source of confusion and not good practice.
1. We also maintain a separate [dev version of our manifests](https://github.com/digitalocean/csi-digitalocean/blob/master/deploy/kubernetes/releases/csi-digitalocean-dev.yaml) that we need to keep in sync with the release copies. In fact, we already have a delta today: [the v1.1.2 release manifest injects the DO API token as secret for the Node plugin](https://github.com/digitalocean/csi-digitalocean/blob/a31f090a350ed24bce07fbd34ed34ae851625fc6/deploy/kubernetes/releases/csi-digitalocean-v1.1.2.yaml#L395-L399) while the dev manifests don't. (The node part does not need a token since it does not speak to the DO API, hence the dev manifests are correct.)

After talking to @varshavaradarajan, a better way to handle this that we came up with is to use the dev manifests as the basis for stamping out release manifests. We modify the dev manifests anyway during development to verify that changes are correct (by, say, running integration tests), so it feels right to use those when cutting a new release.
It's probably also good to change the name from _dev_ to _latest_ then because that's what the now-dev manifests are going to turn into: our latest bleeding edge copy of the manifests that get promoted to release copies at one point.

PR #208, opened 2019-11-11
Use the latest csi driver manifest while bumping up the version for release

varshavaradarajan:
fixes: #206 

This PR: 
- Renames the dev manifest to latest
- Modifies bump-version to copy the new version's manifest from latest (or dev)
- Replaces `do-csi-plugin:dev` with `do-csi-plugin:${NEW_VERSION}`

### DIFF
--- a/deploy/kubernetes/releases/csi-digitalocean-latest.yaml
+++ b/deploy/kubernetes/releases/csi-digitalocean-latest.yaml
@@ -19,6 +19,7 @@
 # example usage: kubectl create -f <this_file>
 #
 
+# This file is only for development use. Do not use in production.
 
 # Install the CSI Driver. This simplifies driver discovery and enables us to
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/releases/csi-digitalocean-latest.yaml
```diff
@@ -19,6 +19,7 @@
 # example usage: kubectl create -f <this_file>
 #
 
+# This file is only for development use. Do not use in production.
```
timoreimann: It might be confusing to users if they find this line in released manifests. Should we omit it or `sed` it out as well during `bump-version`?

varshavaradarajan: Ah, yes, sorry, I didn't commit that. Will do.

PR #210, opened 2019-11-11
Add http health check

nanzhong:
We would like to be able to monitor the availability of the CSI driver. This PR adds a health check endpoint for this.

This is a similar change to https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/293, but simpler since we have full control over how we configuring the driver.

Similar future consideration include:
- tying the DO api health to the actual endpoints that we use

### DIFF
--- a/driver/health.go
+++ b/driver/health.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2018 DigitalOcean
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/digitalocean/godo"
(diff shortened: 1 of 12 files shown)


nanzhong:
🤔 there was an issue with the vendored packages and when I did a tidy + vendored a massive set of changes were pulled in, taking a look to see what's up.

nanzhong:
Figured out the reason for the vendor issues. It was related to some indirect dependencies that is introduced by adding the `healthz` package from `apiserver`. This PR is now updated to with the minimal set of module and vendor changes.

We are running with 1.12 dependencies; I've opened https://github.com/digitalocean/csi-digitalocean/pull/212 to look into updating our dependencies to 1.15.

nanzhong:
@timoreimann would mind taking another look? I've stripped down the health checking to not need new things vendored.

nanzhong:
I will backport this to the other branch PRs after this is thumbsed.

driver/health.go
```diff
@@ -0,0 +1,79 @@
...
+func (c *doHealthChecker) Name() string {
+	return "godo"
+}
+
+func (c *doHealthChecker) Check(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, doHealthTimeout)
+	defer cancel()
+	_, _, err := c.account.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("checking do health: %w", err)
```
timoreimann: nit: wrapping (`%w`) probably not needed.

(This will likely also fail on older versions built with Go <1.13.)

timoreimann:
Conflicts need to be resolved still.

timoreimann:
Oh would be great if you could amend the change log real quick @nanzhong .

PR #218, opened 2019-11-25
Document features

timoreimann:
The change adds a features section outlining the most relevant plugin features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ bumped following the rules below:
 
 ## Features
 
-Below is a list of functionality implemented by the plugin. In general, [CSI features](https://kubernetes-csi.github.io/docs/features.html) implementing a specific aspect of the [specification](https://github.com/container-storage-interface/spec/blob/master/spec.md) are available on any DigitalOcean Kubernetes version for which beta support for the feature is provided.
+Below is a list of functionality implemented by the plugin. In general, [CSI features](https://kubernetes-csi.github.io/docs/features.html) implementing an aspect of the [specification](https://github.com/container-storage-interface/spec/blob/master/spec.md) are available on any DigitalOcean Kubernetes version for which beta support for the feature is provided.
 
-See also the [project examples](/examples/kubernetes) for applicable use cases.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,52 @@ bumped following the rules below:
...
+      [...]
+```
+
+After successful expansion, the _status_ section of the PVC object will reflect the actual capacity.
+
+Important notes:
+
+* Volumes can only be increased in size, not decreased; attempts to do so will lead to an error.
+* Expanding a volume that is larger than the target size will have no effect. The PVC object status section will continue to represent the actual size.
+* Resizing volumes other than through the PVC object (e.g., the DigitalOcean cloud control panel) is not recommended as this cam potentially cause conflicts. Additionally, size updates will not be reflected in the PVC object status field immediately.
```
varshavaradarajan: cam => can

Do we want to mention that eventually, the PVC object will reflect the volume capacity and not the target size in such cases?

README.md
```diff
@@ -16,6 +16,52 @@ bumped following the rules below:
 * New features (such as CSI spec bumps with no breaking changes) will be released as a `MINOR` update.
 * Significant breaking changes makes a `MAJOR` update.
 
+## Features
+
+Below is a list of functionality implemented by the plugin. In general, [CSI features](https://kubernetes-csi.github.io/docs/features.html) implementing a specific aspect of the [specification](https://github.com/container-storage-interface/spec/blob/master/spec.md) are available on any DigitalOcean Kubernetes version for which beta support for the feature is provided.
```
waynr: > a specific aspect of the specification

feels clumsy.

> an aspect of the specification

as an alternative means exactly the same thing and is more concise

README.md
```diff
@@ -16,6 +16,52 @@ bumped following the rules below:
...
+
+See also the [project examples](/examples/kubernetes) for applicable use cases.
```
waynr: How is an "applicable use case" different from an ordinary run of the mill "use case"?

timoreimann:
@waynr @varshavaradarajan comments addressed.

PR #222, opened 2019-11-27
Add Makefile target to check for uncommitted dep changes

timoreimann:
This makes sure changes to the Go modules files and vendor directory are not missed (as it happened in #220).

We also extend the default Makefile target so that Travis CI executes the new target on each run.

Makefile target provided by courtesy of [digitalocean/digitalocean-cloud-controller-manager](https://github.com/digitalocean/digitalocean-cloud-controller-manager/).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ PKG ?= github.com/digitalocean/csi-digitalocean/cmd/do-csi-plugin
 VERSION ?= $(shell cat VERSION)
 DOCKER_REPO ?= digitalocean/do-csi-plugin
 
-all: test
+all: check-unused test
 
 publish: compile build push clean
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,6 +36,12 @@ compile:
 	@echo "==> Building the project"
 	@docker run --rm -it -e GOOS=${OS} -e GOARCH=amd64 -v ${PWD}/:/app -w /app golang:1.13-alpine sh -c 'apk add git && go build -o cmd/do-csi-plugin/${NAME} -ldflags "$(LDFLAGS)" ${PKG}'
 
+.PHONY: check-unused
+check-unused:
```
nanzhong: Could we have this task depend on `vendor` for some DRY?

timoreimann: Very good suggestion, done.

PR #228, opened 2019-12-08
Update --allow-privileged flag usage instructions for kubelet

### LINKED ISSUE #223: Installation notes out of date
### What did you do? (required. The issue will be **closed** when not provided.)
Try to create a Kubernetes cluster following the instructions in this project's README, including using the `--allow-privileged` flag for both controller and worker kubelets. That flag was [removed](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#node) for k8s 1.15 though, so kubelet won't start as a result. 

### What did you expect to happen?
To be able to launch my k8s cluster.

* CSI Version: a82e158764682e57660d7fdbba22cbbc0dbcb12f

* Kubernetes Version: 1.16.3

timoreimann:
[The flag was removed in Kubernetes 1.15](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#node).

Discovered by @aknuds1

Fixes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ Skip secret creation (section 1. in following deployment instructions) when usin
 **Requirements:**
 
 * `--allow-privileged` flag must be set to true for the API server
-* `--allow-privileged` flag must be set to true for the kubelet in Kubernetes 1.14 and below, and not be set otherwise
+* `--allow-privileged` flag must be set to true for the kubelet in Kubernetes 1.14 and below (flag does not exist in later releases)
 * `--feature-gates=KubeletPluginsWatcher=true,CSINodeInfo=true,CSIDriverRegistry=true` feature gate flags must be set to true for both the API server and the kubelet
 * Mount Propagation needs to be enabled. If you use Docker, the Docker daemon of the cluster nodes must allow shared mounts.
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,8 @@ Skip secret creation (section 1. in following deployment instructions) when usin
 
 **Requirements:**
 
-* `--allow-privileged` flag must be set to true for both the API server and the kubelet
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,7 +124,8 @@ with the CSI driver pre-installed and no further steps are required.
 
 **Requirements:**
 
-* `--allow-privileged` flag must be set to true for both the API server and the kubelet
+* `--allow-privileged` flag must be set to true for the API server
+* `--allow-privileged` flag must be set to true for the kubelet in Kubernetes 1.14 and below, and not be set otherwise
```
adamwg: `and not be set otherwise` makes it sound a bit like setting it will cause problems specifically for our CSI driver, whereas actually the issue is that it doesn't exist. I would probably just call out here that the flag is removed in later releases, something like `... in Kubernetes 1.14 and below (flag does not exist in later releases)`.

timoreimann: Good suggestion, done. I might adjust a wording a bit more if it turns out that the API server flag might not be needed, either generally or in specific versions. (See main discussion on this PR.)

timoreimann:
@aknuds1 good question -- the [api-server CLI reference](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) says this about the parameter:

```
--allow-privileged

    If true, allow privileged containers. [default=false]
```

I wonder if the usage of PSP implies this parameter somehow and provides an additional, declarative gate. We should test what the behavior is like if we also drop the API server parameter and _not_ use PSP.

aknuds1:
@timoreimann That would be great, thanks!

timoreimann:
@aknuds1 do you happen to have a working PSP definition available that I could test with?

PR #229, opened 2019-12-09
Refactor waitAction

timoreimann:
This change refactors `waitAction` to use `wait.PollUntil` from the upstream apimachinery library to hide the underlying polling infrastructure and simplify the implementation.

We also exit the method if the context got cancelled while waiting for a response from the volume API. Previously, we would iterate through the for loop once again and possibly do another unnecessary API interaction if the ticker was ready again by that time.

Finally, we add a unit test (which required making the wait action timeout injectable).

The change passes both our local integration and upstream e2e tests.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


driver/controller.go
```diff
@@ -950,34 +951,35 @@ func (d *Driver) waitAction(ctx context.Context, volumeId string, actionId int)
...
 	})
 
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	// This timeout should not strike given all sidecars use a timeout that is
+	// lower (which should always be the case). Using this as a fail-safe
+	// mechanism in case of a bug or misconfiguration.
+	ctx, cancel := context.WithTimeout(ctx, d.waitActionTimeout)
 	defer cancel()
 
-	// TODO(arslan): use backoff in the future
```
timoreimann: Note to reviewers: I did not carry this TODO intentionally since backoffs for timeouts are implemented by the sidecars, and other kinds of errors we occasionally run into do not seem to warrant an exponential backoff as of today (e.g., temporary 4xx responses returned from the API that disappear on the next API call).

PR #233, opened 2019-12-10
Return error when fetching the snapshot fails

timoreimann:
We previously continued processing when fetching the snapshot failed for errors other than 404.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -169,6 +169,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			if resp != nil && resp.StatusCode == http.StatusNotFound {
 				return nil, status.Errorf(codes.NotFound, "snapshot %q not found", snapshotID)
 			}
+			return nil, err
 		}
 
 		ll.WithField("snapshot_id", snapshotID).Info("using snapshot as volume source")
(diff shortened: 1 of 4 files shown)


driver/controller.go
```diff
@@ -169,6 +169,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			if resp != nil && resp.StatusCode == http.StatusNotFound {
 				return nil, status.Errorf(codes.NotFound, "snapshot %q not found", snapshotID)
 			}
+			return nil, err
```
timoreimann: We do not seem to return a proper gRPC status code in similar code paths either, so I copied the pattern. Not sure if good or not, my expectation would be that gRPC maps this to a generic "internal error" or similar.

@adamwg you may know this one better than me.

adamwg: I believe the go gRPC framework wraps non-gRPC errors from RPC handlers using the `Unknown` status code. I always prefer to return explicit gRPC errors with appropriate codes from handlers, but gRPC is OK with it either way.

Makes sense to follow the general pattern here, and if we want to return explicit `Internal`s everywhere we can do that in a separate RP.

timoreimann:
I added `TestCreateVolume` and covered two exemplary cases, one of them addressing the code modified. Let me know what you think.

driver/controller.go
timoreimann: Makes sense to me -- will keep that in mind for a follow up PR.

PR #235, opened 2019-12-10
Support checkLimit for multiple pages

timoreimann:
This change ~implements paging~ uses the `Meta.Total` field from the godo response when listing volumes for account limit checking, thereby supporting limits beyond the size of a single page.

We also refactor `checkLimits` to not return gRPC codes but leave that to the calling function. This also allows us to distinguish between true errors and limit violations, and better handle each case.


adamwg:
Also wanted to note: the volumes API in particular also returns a `meta` field in the response, which explicitly says the total number of volumes the account has:

```
  "meta": {
    "total": 8316
  },
```

I think godo ignores this, but we could integrate it into godo and leverage it here rather than relying on the pagination data.

jcodybaker:
Reviewed.  No concerns other than those already raised by AWG.

timoreimann:
@adamwg I like the idea to use the page count in order to determine the number of volumes until the `meta` field is available in godo.

Looking more closely at things, however, it looks like the number of pages isn't directly available in godo but only through [the `Last` reference](https://github.com/digitalocean/csi-digitalocean/blob/39cd53b2791b03fc23db6ccfd7dc6912084aca2a/vendor/github.com/digitalocean/godo/links.go#L19) which seems to be a URL. The functionality godo uses to parse page numbers from the links seems non-exported.

I'd rather not implement / copy the parsing logic in(to) our repo, which makes an upstream contribution to godo in order to expose the `meta` field more desirable to me right now. But please let me know if I missed something.

(FWIW, there's no super urge to get this PR in, so we can wait a bit for any godo PR to get merged and released.)

timoreimann:
Submitted digitalocean/godo#286.

timoreimann:
my godo contribution was merged and released. This PR is now using godo 1.29.0 and leverages the `Meta.Total` field.

@adamwg @jcodybaker would you mind taking a look again when you have a minute?

timoreimann:
@jcodybaker 

>Have you verified that list volumes honors and correctly implements the per-page=1 option?

That does seem to be the case:

```shell
$ curl -sfX GET -H "Content-Type: application/json" -H "Authorization: Bearer <REDACTED>" "https://api.digitalocean.com/v2/volumes?per_page=1" | jq '. | {meta: .meta, numVolumes: .volumes | length}'

{
  "meta": {
    "total": 14
  },
  "numVolumes": 1
}
```

jcodybaker:
Nice :shipit:

PR #238, opened 2019-12-11
Implement graceful shutdown

timoreimann:
This change implements graceful shutdown so that a proper driver termination leads to an exit code of zero.

Specifically, we handle signals, and let the HTTP and gRPC servers finish requests gracefully.

We also remove the `Stop()` function in favor a context-based approach.

### DIFF
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/digitalocean/godo"
 	"github.com/kubernetes-csi/csi-test/pkg/sanity"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


driver/driver_test.go
```diff
@@ -81,9 +81,16 @@ func TestDriverSuite(t *testing.T) {
...
 	}
-	defer driver.Stop()
 
-	go driver.Run()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		err := driver.Run(ctx)
+		if err != nil {
```
adamwg: If we want to check this error reliably, we should probably call `cancel` after `sanity.Test` returns, and use a WaitGroup or channel to wait for this goroutine to shut down before the test ends. As it's written currently, I suspect the test will be finished before this line is executed, since the `defer` will execute right at the end of the test function.

timoreimann: Ah, good catch.

I went with an `errgroup`. Let me know if that makes sense to you.

adamwg: Yep, makes sense to me :+1:

PR #244, opened 2019-12-16
Fix and improve logging

timoreimann:
This change applies a number of improvements to the way we log:

- Rename "node_id" field set on controller initialization to "host_id" to avoid confusion with the "node_id" we set in various gRPC endpoint implementations: the former specifies we host we run on while the latter refers to the node ID passed in from the request. The two may diverge, especially on the controller.
- Pass existing logrus entry to `waitAction` to expand fields on.
- Log methods called and returned more consistently.
- Rename the log variable `ll` to the more readable `log`.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -349,7 +349,7 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 
 	if action != nil {
 		ll.Info("waiting until volume is attached")
-		if err := d.waitAction(ctx, req.VolumeId, action.ID); err != nil {
+		if err := d.waitAction(ctx, ll, req.VolumeId, action.ID); err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


driver/controller.go
```diff
@@ -839,7 +837,14 @@ func (d *Driver) ControllerExpandVolume(ctx context.Context, req *csi.Controller
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "ControllerExpandVolume could not retrieve existing volume: %v", err)
 	}
-	ll.Info("extracting storage from requested capacity range")
+
+	ll := d.log.WithFields(logrus.Fields{
```
snormore: Can we call this `log` instead of `ll`? 😅

timoreimann: Good idea -- massively renamed this everywhere. :)

PR #247, opened 2019-12-17
Apply improvements to logging

timoreimann:
Includes logging missing volume on `ControllerUnpublishVolume` and further improvements.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -168,7 +168,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		_, resp, err := d.snapshots.Get(ctx, snapshotID)
 		if err != nil {
 			if resp != nil && resp.StatusCode == http.StatusNotFound {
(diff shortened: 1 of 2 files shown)


timoreimann:
@nicktate that's a great suggestion. I applied further improvements to our logging across the board.

driver/controller.go
```diff
@@ -425,18 +424,17 @@ func (d *Driver) ControllerUnpublishVolume(ctx context.Context, req *csi.Control
 					log.WithFields(logrus.Fields{
 						"error": err,
 						"resp":  resp,
-					}).Warn("assuming volume is detached already")
+					}).Warn("assuming volume is detached because of error response")
```
jcodybaker: 🤷‍♂ - This might be more clear if it said something like "assuming volume is detached; api reports 'Attachment not found'".  Good with it as is though.

timoreimann: My thinking here was that the `error` field we include in the log output includes the reason which our log message points at. The intention was to avoid repetition.

jcodybaker: 👍

PR #248, opened 2019-12-27
Add support for running upstream storage end-to-end tests

timoreimann:
This change adds tooling and scaffolding necessary to validate the DigitalOcean CSI driver against a number of supported Kubernetes releases using the upstream storage end-to-end tests. It is built on top of Go's testing package and faciliates creating DOKS clusters on-the-fly to deploy a driver-under-test into.

See the included README file for details.

The change does not integrate the tests into CI yet. A future PR will address this final piece.

Suggested order of files to review:

1. `README.md`
1. `Dockerfile`
1. `testdrivers/1.16.yaml`
1. `run-versioned-e2e-tests.sh`
1. `e2e.sh`
1. `e2e_test.go`
1. `*_test.go`

Example output:

```shell
$ ./e2e.sh -driver-image timoreimann/do-csi-plugin:dev 1.16
=== RUN   TestE2E
=== RUN   TestE2E/1.16
Created cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt) (response code: 201)
Waiting for cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt) to become running
Current status of cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt): provisioning (message: provisioning)
Current status of cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt): provisioning (message: provisioning)
[...]
Current status of cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt): provisioning (message: provisioning)
Cluster 8b5a36fa-bbff-4f1d-b749-c6daa233be31 (csi-e2e-1-16-test-gkwzt) is running
Deploying a dev version of the CSI driver to context do-fra1-csi-e2e-1-16-test-gkwzt.
[...]
Running parallel tests
Running Suite: Kubernetes e2e suite
===================================
Random Seed: 1577484230 - Will randomize all specs
Will run 4896 specs

Running in parallel across 10 nodes

Dec 27 22:03:52.763: INFO: >>> kubeConfig: /root/.kube/config
Dec 27 22:03:52.774: INFO: Waiting up to 30m0s for all (but 0) nodes to be schedulable
[...]

Issue #209, opened 2019-11-11
Cannot install Rook using DO PVs -  /var/lib/kubelet/plugins/dobs.csi.digitalocean.com/csi.sock: connect: connection refused


MasterNayru:
### What did you do? (required. The issue will be **closed** when not provided.)
I am trying to install Rook on my Kubernetes cluster using DigitalOcean PVCs as the backing storage/mon disks. The manifests and setup instructions I am using are pretty much straight from the getting started docs on Rook's site, but it seems that Rook is having its connections to the DigitalOcean CSI Unix socket blocked with the following events on the Rook OSD prepare pods:
```
Events:
  Type     Reason           Age                    From                  Message
  ----     ------           ----                   ----                  -------
  Warning  FailedMount      12m (x520 over 19h)    kubelet, k8sworker02  (combined from similar events): Unable to attach or mount volumes: unmounted volumes=[set1-0-data-fch6l], unattached volumes=[rook-data udev rook-binaries rook-ceph-osd-token-wjd6x set1-0-data-fch6l-bridge set1-0-data-fch6l ceph-conf-emptydir rook-ceph-log devices]: timed out waiting for the condition
  Warning  FailedMapVolume  5m58s (x399 over 20h)  kubelet, k8sworker02  MapVolume.SetUp failed for volume "pvc-2578a0e1-ba2d-4144-86a2-c2d03ad73642" : kubernetes.io/csi: blockMapper.stageVolumeForBlock failed to check STAGE_UNSTAGE_VOLUME capability: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial unix /var/lib/kubelet/plugins/dobs.csi.digitalocean.com/csi.sock: connect: connection refused"
  Warning  FailedMapVolume  114s (x203 over 20h)   kubelet, k8sworker02  MapVolume.SetUp failed for volume "pvc-2578a0e1-ba2d-4144-86a2-c2d03ad73642" : kubernetes.io/csi: blockMapper.stageVolumeForBlock failed: rpc error: code = Unavailable desc = transport is closing
```

The volumes themselves have been created, and bound to PVCs correctly as shown here, so I assume that I have set up the driver correctly:
```
 ~ $ kubectl get pvc -n rook-ceph     
NAME                STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS       AGE
[...]

PR #249, opened 2020-01-03
Raw block volume mode support

nicktate:
#### What does this pull request accomplish

* Adds `VolumeMode: Block` support 
* Parallelizes `test/integration_test.go`
* Adds `Block` volume tests to integration tests

#### Additional Info 

Here is a high level summary of the changes to the driver:

* `controller.go/validateCapabilities` - Needs to accept `Block` access mode in the capabilities check
* `node.go/NodeStageVolume` - Updated to a `noop` for block volumes because we bind mount the absolute device path directly
* `node.go/NodePublishVolume` - Checks the `AccessMode` and triggers `nodePublishVolumeForFileSystem` or `nodePublishVolumeForBlock` respectively. 
  * The main difference is the source path for `nodePublishVolumeForBlock` is the absolute path of the device on disk
* `mounter.go/Mount` - Updated to handle an empty `fsType` being passed in the case of `Block`
* `node.go/NodeGetVolumeStats` - Checks to determine if it is a block device. If so, it uses `blockdev` to determine the `totalCapacity` of the device (no other information can be collected)

Here is a high level summary of the changes to the integration tests:

* Supports running tests in parallel
* Generate resource names based off of the test name
* Add table driven tests, one per volume mode: `Filesystem` and `Block`
* Clean up all resources on test exit

#### Test Results

* Integration Tests
  * 1.16 - https://gist.github.com/nicktate/b0e6de645afea812058399eb691246dd
  * 1.13 - https://gist.github.com/nicktate/5695a3fefce623cc607d17c4d5c74896
* E2E Tests
  * https://gist.github.com/nicktate/108b91295d9c7db7bf02d579682bd2b5
  * See https://github.com/digitalocean/csi-digitalocean/pull/248 for more info

### DIFF
--- a/driver/mounter.go
+++ b/driver/mounter.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -47,7 +49,7 @@ type volumeStatistics struct {
 
 const (
 	// blkidExitStatusNoIdentifiers defines the exit code returned from blkid indicating that no devices have been found. See http://www.polarhome.com/service/man/?qf=blkid&tf=2&of=Alpinelinux for details.
-	blkidExitStatusNoIdentifiers  = 2
+	blkidExitStatusNoIdentifiers = 2
 )
 
 // Mounter is responsible for formatting and mounting volumes
@@ -75,6 +77,9 @@ type Mounter interface {
 	// GetStatistics returns capacity-related volume statistics for the given
 	// volume path.
 	GetStatistics(volumePath string) (volumeStatistics, error)
+
+	// IsBlockDevice checks whether the device at the path is a block device
+	IsBlockDevice(volumePath string) (bool, error)
 }
 
 // TODO(arslan): this is Linux only for now. Refactor this into a package with
@@ -135,10 +140,6 @@ func (m *mounter) Mount(source, target, fsType string, opts ...string) error {
 	mountCmd := "mount"
 	mountArgs := []string{}
 
-	if fsType == "" {
-		return errors.New("fs type is not specified for mounting the volume")
-	}
-
 	if source == "" {
 		return errors.New("source is not specified for mounting the volume")
 	}
@@ -147,7 +148,29 @@ func (m *mounter) Mount(source, target, fsType string, opts ...string) error {
 		return errors.New("target is not specified for mounting the volume")
 	}
 
-	mountArgs = append(mountArgs, "-t", fsType)
+	// This is a raw block device mount. Create the mount point as a file
+	// since bind mount device node requires it to be a file
+	if fsType == "" {
+		// create directory for target, os.Mkdirall is noop if directory exists
+		err := os.MkdirAll(filepath.Dir(target), 0750)
+		if err != nil {
+			return fmt.Errorf("failed to create target directory for raw block bind mount: %v", err)
+		}
+
+		file, err := os.OpenFile(target, os.O_CREATE, 0660)
+		if err != nil {
+			return fmt.Errorf("failed to create target file for raw block bind mount: %v", err)
+		}
+		file.Close()
+	} else {
+		mountArgs = append(mountArgs, "-t", fsType)
+
+		// create target, os.Mkdirall is noop if directory exists
+		err := os.MkdirAll(target, 0750)
+		if err != nil {
+			return err
+		}
+	}
 
 	if len(opts) > 0 {
 		mountArgs = append(mountArgs, "-o", strings.Join(opts, ","))
@@ -155,12 +178,6 @@ func (m *mounter) Mount(source, target, fsType string, opts ...string) error {
 
 	mountArgs = append(mountArgs, source)
 	mountArgs = append(mountArgs, target)
-
-	// create target, os.Mkdirall is noop if it exists
-	err := os.MkdirAll(target, 0750)
-	if err != nil {
-		return err
-	}
 
 	m.log.WithFields(logrus.Fields{
 		"cmd":  mountCmd,
@@ -221,20 +238,20 @@ func (m *mounter) IsFormatted(source string) (bool, error) {
 
 	exitCode := 0
 	cmd := exec.Command(blkidCmd, blkidArgs...)
-  err = cmd.Run()
-  if err != nil {
-		exitError, ok := err.(*exec.ExitError);
-    if !ok {
+	err = cmd.Run()
+	if err != nil {
+		exitError, ok := err.(*exec.ExitError)
+		if !ok {
 			return false, fmt.Errorf("checking formatting failed: %v cmd: %q, args: %q", err, blkidCmd, blkidArgs)
 		}
 		ws := exitError.Sys().(syscall.WaitStatus)
 		exitCode = ws.ExitStatus()
-		if (exitCode == blkidExitStatusNoIdentifiers) {
+		if exitCode == blkidExitStatusNoIdentifiers {
 			return false, nil
 		} else {
 			return false, fmt.Errorf("checking formatting failed: %v cmd: %q, args: %q", err, blkidCmd, blkidArgs)
 		}
-  }
+	}
 
 	return true, nil
 }
@@ -299,9 +316,31 @@ func (m *mounter) IsMounted(target string) (bool, error) {
 }
 
 func (m *mounter) GetStatistics(volumePath string) (volumeStatistics, error) {
+	isBlock, err := m.IsBlockDevice(volumePath)
+	if err != nil {
+		return volumeStatistics{}, fmt.Errorf("failed to determine if volume %s is block device: %v", volumePath, err)
+	}
+
+	if isBlock {
+		// See http://man7.org/linux/man-pages/man8/blockdev.8.html for details
+		output, err := exec.Command("blockdev", "getsize64", volumePath).CombinedOutput()
+		if err != nil {
+			return volumeStatistics{}, fmt.Errorf("error when getting size of block volume at path %s: output: %s, err: %v", volumePath, string(output), err)
+		}
+		strOut := strings.TrimSpace(string(output))
+		gotSizeBytes, err := strconv.ParseInt(strOut, 10, 64)
+		if err != nil {
+			return volumeStatistics{}, fmt.Errorf("failed to parse size %s into int", strOut)
+		}
+
+		return volumeStatistics{
+			totalBytes: gotSizeBytes,
+		}, nil
+	}
+
 	var statfs unix.Statfs_t
 	// See http://man7.org/linux/man-pages/man2/statfs.2.html for details.
-	err := unix.Statfs(volumePath, &statfs)
+	err = unix.Statfs(volumePath, &statfs)
 	if err != nil {
 		return volumeStatistics{}, err
 	}
@@ -317,4 +356,14 @@ func (m *mounter) GetStatistics(volumePath string) (volumeStatistics, error) {
 	}
 
 	return volStats, nil
+}
+
+func (m *mounter) IsBlockDevice(devicePath string) (bool, error) {
+	var stat unix.Stat_t
+	err := unix.Stat(devicePath, &stat)
+	if err != nil {
+		return false, err
+	}
+
+	return (stat.Mode & unix.S_IFMT) == unix.S_IFBLK, nil
 }
--- a/driver/node.go
+++ b/driver/node.go
@@ -26,7 +26,9 @@ package driver
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/sirupsen/logrus"
@@ -42,6 +44,9 @@ const (
 
 	// See: https://www.digitalocean.com/docs/volumes/overview/#limits
 	maxVolumesPerNode = 7
+
+	volumeModeBlock      = "block"
+	volumeModeFilesystem = "filesystem"
 )
 
 var (
@@ -87,7 +92,14 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		volumeName = volName
 	}
 
-	source := getDiskSource(volumeName)
+	// If it is a block volume, we do nothing for stage volume
+	// because we bind mount the absolute device path to a file
+	switch req.VolumeCapability.GetAccessType().(type) {
+	case *csi.VolumeCapability_Block:
+		return &csi.NodeStageVolumeResponse{}, nil
+	}
+
+	source := getDeviceByIDPath(volumeName)
 	target := req.StagingTargetPath
 
 	mnt := req.VolumeCapability.GetMount()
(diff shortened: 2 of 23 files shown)


driver/mounter.go
```diff
@@ -299,9 +318,31 @@ func (m *mounter) IsMounted(target string) (bool, error) {
 }
 
 func (m *mounter) GetStatistics(volumePath string) (volumeStatistics, error) {
+	isBlock, err := m.IsBlockDevice(volumePath)
+	if err != nil {
+		return volumeStatistics{}, err
```
timoreimann: Shall we annotate the error?

timoreimann:
I forgot to mention that I haven't reviewed the extended integration tests yet; will do that in a second round.

@nicktate btw thanks for the excellent PR description / overview -- appreciated the quick overview I was able to get out of that. 👏

timoreimann:
The [spec describes](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodeexpandvolume) how block the `NodeExpandVolume` call may handle block mode:

```
  // Volume capability describing how the CO intends to use this volume.
  // This allows SP to determine if volume is being used as a block
  // device or mounted file system. For example - if volume is being
  // used as a block device the SP MAY choose to skip expanding the
  // filesystem in NodeExpandVolume implementation but still perform
  // rest of the housekeeping needed for expanding the volume. If
  // volume_capability is omitted the SP MAY determine
  // access_type from given volume_path for the volume and perform
  // node expansion. This is an OPTIONAL field.
  VolumeCapability volume_capability = 5;```

This raises two questions for me:

1. Should we skip expanding in case of block mode access?
2. (If yes to 1:) Should we implement access type detection if `volume_capability` is omitted?

File: test/kubernetes/integration_test.go [hunk 2]
```diff
@@ -132,414 +153,481 @@ func TestPod_Single_Volume(t *testing.T) {
...
+
+func TestPod_Single_Volume(t *testing.T) {
+	t.Parallel()
+	appName := testAppName(t)
+	volumeName := "my-do-volume"
+	claimName := testClaimName(t)
+
+	tt := []struct {
+		pod func() *v1.Pod
+		pvc func() *v1.PersistentVolumeClaim
```
timoreimann: I remember we discussed that the integration tests will only be useful for testing Kubernetes < 1.14 going forward because newer releases support running the upstream end-to-end tests (that ship with an entire suite of access mode-related tests). What we likely want at some point is introduce a flag to run all integration tests for older releases (1.13 while still supported by us, potentially even older ones) and none for newer releases.

I understand the function returning a PVC is to support testing for filesystem and block mode, respectively, here and for most (all?) test we have build so far. Unfortunately, this would make it difficult to run the tests on 1.13 or below because [block mode turned beta in Kubernetes 1.14 only](https://kubernetes-csi.github.io/docs/raw-block.html); meaning that the block mode-related tests would all fail on older releases.

Thinking about how we can resolve this, I can see two possibilities:

1. Test the access mode in a single, dedicated test only and revert the pre-existing tests to their prior state. That way, we can easily turn tests on and off as needed per tested Kubernetes release.
1. Do not test the access mode in our integration tests at all but rely on the upstream tests only. Merging the feature without any tests is not ideal, however, so for this scenario we'd probably want to wait for #248 first so that we can enable the block testing capability in the upstream tests.

The first option allows us to move forward without depending on #248, at the price of a bit more tech debt (as we eventually will be able to remove our access mode integration tests entirely in favor of the upstream ones; at least, I assume so). #248 seems close to completion, but you never know, so maybe option 1. seems slightly preferable.

Let me know what you think.

timoreimann:
Left a general comment/question regarding the direction of our integration tests, which I think we should answer first before going into details.

test/kubernetes/integration_test.go
nicktate: Good catch on the version support :+1: 

I think #248 is definitely the direction we will move towards but I think `1` or a variation of it would be better in the short term to reduce the risk of this PR getting blocked. 

I'm fine either creating a separate test as you suggested in number `1` or simply calculating the `kubernetes_version` of the cluster in test during `setup` and then having a minimum version gate for the table driven test configuration. I think either way it would be smart to create a version gate for the block tests.

I'm personally leaning towards adding the gate into the updated table driven tests for the enhanced test cleanup functionality and naming. I think if they were staying long term it would be the right decision, but since these tests also have a relatively short EOL I would be fine with reverting them to prior state as well.

driver/mounter.go
```diff
@@ -147,7 +148,31 @@ func (m *mounter) Mount(source, target, fsType string, opts ...string) error {
 		return errors.New("target is not specified for mounting the volume")
 	}
 
-	mountArgs = append(mountArgs, "-t", fsType)
+	// This is a raw block device mount. Create the mount point as a file
+	// since bind mount device node requires it to be a file
```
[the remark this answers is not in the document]
nicktate: Trial and error and then found the corresponding man page (https://linux.die.net/man/8/mount) information about it which was extremely vague:

```
mount --bind olddir newdir

or shortoption
mount -B olddir newdir

or fstab entry is:
/olddir /newdir none bind

After this call the same contents is accessible in two places. One can also remount a single file (on a single file).
```

Specifically: 
> One can also remount a single file (on a single file).

Given that the absolute device path is a block device file, we correspondingly have to bind mount that onto another file.

driver/node.go
```diff
@@ -447,8 +464,113 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	stat, err := os.Lstat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", fmt.Errorf("device does not exist at path %q", path)
+		}
+		return "", fmt.Errorf("failed to Lstat %q: %v", path, err)
+	}
+
+	if stat.Mode()&os.ModeSymlink != os.ModeSymlink {
+		return "", fmt.Errorf("device found, but was not a symlink %q", path)
```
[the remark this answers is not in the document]
adamwg: Agreed - I would call `EvalSymlinks` to start with. Realistically, the `/dev/disk/by-id/...` will probably always be a symlink, but it shouldn't break anything if it's the actual device or a hardlink.

driver/mounter.go
```diff
@@ -147,7 +148,31 @@ func (m *mounter) Mount(source, target, fsType string, opts ...string) error {
...
+	if fsType == "" {
+		// create directory for target, os.Mkdirall is noop if directory exists
+		err := os.MkdirAll(filepath.Dir(target), 0750)
+		if err != nil {
+			return fmt.Errorf("failed to create target directory for raw block bind mount: %v", err)
+		}
+
+		file, err := os.OpenFile(target, os.O_CREATE, 0666)
+		if file != nil {
+			file.Close()
```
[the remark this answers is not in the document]
nicktate: If the `err` is non-nil, then `file` will always be `nil`. 

```
https://golang.org/src/os/file.go
func OpenFile(name string, flag int, perm FileMode) (*File, error) {
	testlog.Open(name)
	f, err := openFileNolog(name, flag, perm)
	if err != nil {
		return nil, err
	}
	f.appendMode = flag&O_APPEND != 0

	return f, nil
}
```

I opted to check `file != nil` because it read more explicitly since we always want the file closed if we found it since we are only concerned of its existence. It could also be written as follows if you think it is better in terms of readibility:

```
...
if err != nil {
	return fmt.Errorf("failed to create target file for raw block bind mount: %v", err)
}
file.Close()
...
```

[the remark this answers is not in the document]
timoreimann: I personally prefer the style to check the error first because it seems more Go idiomatic and saves us one `if`. Practically, it shouldn't make any difference, so I'm happy to let you make the call.

nicktate:
@adamwg / @timoreimann I agree with you both on the `NodeExpandVolume` (lean towards *not* automatically expanding in the case Block mode access). 

https://github.com/kubernetes/kubernetes/blob/4c50ee993c82c6852eb3b3aa8dfa8ecc4bcfe330/pkg/util/resizefs/resizefs_linux.go#L49 ignores resizing of the device if it is not formatted. I hadn't considered the case of the user requesting block and formatting after the fact. 

I will `noop` if Block is in volume capability, but if it omitted, I think it is fine to attempt the resize since it will be internally handled.

test/kubernetes/integration_test.go
timoreimann: Gating on top of your improvements sounds fine for me as well. 👍

driver/mounter.go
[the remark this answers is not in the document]
timoreimann: 👍Thanks (and good job discovering this)

timoreimann:
@nicktate 

>I will `noop` if Block is in volume capability, but if it omitted, I think it is fine to attempt the resize since it will be internally handled.

That sounds good to me. 👍

FWIW, the `volumeMode` field is immutable, so users wouldn't be able to change the mode through the Kubernetes API. They'd have to access the volume directly, at which points all bets are off.

driver/node.go
```diff
@@ -447,8 +472,101 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	if mnt.FsType != "" {
+		fsType = mnt.FsType
+	}
+
+	mounted, err := d.mounter.IsMounted(target)
+	if err != nil {
+		return err
+	}
+
+	log.WithFields(logrus.Fields{
```
timoreimann: `log.WithFields()` needs to be assigned to `log` again or otherwise the extra fields won't show up in the `log.Info()` statements below.

driver/node.go
```diff
@@ -447,8 +472,101 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	}
+
+	target := req.TargetPath
+
+	mounted, err := d.mounter.IsMounted(target)
+	if err != nil {
+		return err
+	}
+
+	log.WithFields(logrus.Fields{
```
timoreimann: `log = log.WithFields(...)` here too.

driver/node.go
```diff
@@ -424,6 +440,15 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
 	})
 	log.Info("node expand volume called")
 
+	log.Infof("volume cap: +v\n", req.GetVolumeCapability())
```
timoreimann: Was this for debugging? Shall we drop this?

driver/node.go
```diff
@@ -447,8 +472,101 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+		log.Info("volume is already mounted")
+	}
+
+	return nil
+}
+
+func (d *Driver) nodePublishVolumeForBlock(req *csi.NodePublishVolumeRequest, mountOptions []string, log *logrus.Entry) error {
+	var ok bool
+	var volumeName string
+	if volumeName, ok = req.GetPublishContext()[d.publishInfoVolumeName]; !ok {
```
timoreimann: Sorry I didn't see this on the first pass, but if we move the`if` to the next line like

```go
volumeName, ok = req.GetPublishContext()[d.publishInfoVolumeName]
if !ok {
  [...]
}
```

then we don't need the preceding variable declarations for `ok` and `volumeName`.

test/kubernetes/deploy/deploy.sh
```diff
@@ -57,4 +57,4 @@ kustomize build . --load_restrictor none | kubectl apply -f -
 # Wait for the deployment to complete.
 kubectl -n kube-system wait --timeout=5m --for=condition=Ready pod -l app=csi-do-controller-dev
 kubectl -n kube-system wait --timeout=5m --for=condition=Ready pod -l app=csi-do-node-dev
-kubectl -n kube-system get all
+kubectl -n kube-system get all
```
timoreimann: nit: newline missing

driver/node.go
nicktate: :+1:, I think that is a better approach

driver/node.go
nicktate: Sorry, looks like I didn't save in the editor before the last `git add` 🤦‍♂  I'll update shortly

driver/node.go
nicktate: 👍

nicktate:
@timoreimann Most comments have been addressed with exception of test updates & readme/changelog. I will be following up once I have re-ran integration / e2e tests.

nicktate:
@timoreimann / @adamwg This should be good for a final review. I believe all comments have been addressed and I've re-ran e2e and upstream tests and posted the new results in the description.

I do have some interesting findings regarding the expand volume. I updated the logic per the spec and check volume capabilities if it exists, and in the case of block volume attempt to `noop` a filesystem resize. In effect, if a user formatted their raw block volume and then expanded the persistent volume, we would also attempt to expand the filesystem because we cannot discern it is block access mode.

In testing the `VolumeCapability` argument is always nil and after digging into the source I have found it is not passed to the corresponding calls:

* https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_client.go#L290-L295
* https://github.com/kubernetes-csi/external-resizer/blob/cbc2ca90732be32b6bc1bcfa2ce48519028086b6/pkg/csi/client.go#L124-L128

I posed the question in the community CSI channel, so you can follow any updates here: https://kubernetes.slack.com/archives/C8EJ01Z46/p1578414177000600.

nicktate:
I've also tested against https://github.com/digitalocean/csi-digitalocean/issues/209 and it seems to be working as expected:

```
╰─ k get po
NAME                                                            READY   STATUS      RESTARTS   AGE
csi-cephfsplugin-88krf                                          3/3     Running     0          4m35s
csi-cephfsplugin-lbjd8                                          3/3     Running     0          4m35s
csi-cephfsplugin-provisioner-86d9db5b7b-l52pg                   4/4     Running     0          4m35s
csi-cephfsplugin-provisioner-86d9db5b7b-tnk8h                   4/4     Running     0          4m35s
csi-cephfsplugin-wlmb7                                          3/3     Running     0          4m35s
csi-rbdplugin-6gzfg                                             3/3     Running     0          4m36s
csi-rbdplugin-cg2zj                                             3/3     Running     0          4m36s
csi-rbdplugin-gqkb6                                             3/3     Running     0          4m36s
csi-rbdplugin-provisioner-76bb56764b-8cx4t                      5/5     Running     0          4m36s
csi-rbdplugin-provisioner-76bb56764b-9crmp                      5/5     Running     0          4m36s
rook-ceph-crashcollector-pool-3yiqj6yja-kxkn-84bc647d4b-btq8j   1/1     Running     0          2m55s
rook-ceph-crashcollector-pool-3yiqj6yja-kzep-6c7bcd8686-mmrfj   1/1     Running     0          90s
rook-ceph-crashcollector-pool-3yiqj6yja-kzes-56cbb64c4f-9dld9   1/1     Running     0          3m14s
rook-ceph-mgr-a-6c7497f4d6-k5j9f                                1/1     Running     0          2m32s
rook-ceph-mon-a-856565757b-sk4t2                                1/1     Running     0          3m34s
rook-ceph-mon-b-5bf967cfb5-gxhnz                                1/1     Running     0          3m14s
rook-ceph-mon-c-6cd5c4bbc4-q9cs2                                1/1     Running     0          2m55s
rook-ceph-operator-f978b8565-j45jq                              1/1     Running     0          5m33s
rook-ceph-osd-0-7bf8956b67-5tczx                                1/1     Running     0          90s
rook-ceph-osd-1-959f64d47-79mxw                                 1/1     Running     0          86s
rook-ceph-osd-2-6df95cd54-tglwn                                 1/1     Running     0          84s
rook-ceph-osd-prepare-set1-0-data-7jgmd-2cwkr                   0/1     Completed   0          2m11s
rook-ceph-osd-prepare-set1-1-data-lc9hv-2ngbw                   0/1     Completed   0          2m11s
rook-ceph-osd-prepare-set1-2-data-bvkk9-55mg6                   0/1     Completed   0          2m11s
rook-discover-h5t7z                                             1/1     Running     0          5m13s
rook-discover-lndkl                                             1/1     Running     0          5m13s
rook-discover-lqdcc                                             1/1     Running     0          5m13s
```

PR #251, opened 2020-01-06
Document all supported Kubernetes versions

timoreimann:
This change explicitly mentions all Kubernetes releases we officially support with our CSI driver. It seems preferable to the previous style which leaves uncertainty as to whether the most recently published Kubernetes release is already supported (e.g., it is not clear if our CSI drivers supports 1.17 today).

We also drop the `dev` row as it is fairly uninteresting to non-developers and is easy to forget to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,61 +66,17 @@ Volumes can be transferred across clusters. The exact steps are outlined in [our
 
 ### Kubernetes Compatibility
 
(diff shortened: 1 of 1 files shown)


timoreimann:
@adamwg great point -- I simplified the table accordingly. PTAL.

PR #259, opened 2020-01-08
Prepare release v1.0.4

timoreimann:


deploy/kubernetes/releases/csi-digitalocean-v1.0.4.yaml
```diff
@@ -0,0 +1,526 @@
+# Copyright 2018 DigitalOcean
```
jcodybaker: Should we bump this copyright date in the template?

timoreimann: Yeah, I guess we should do that. How about doing one big update in a dedicated, follow-up PR though? Or would you prefer updating just the template in this one?

jcodybaker: Yeah, I don't have strong feelings on it.  It's open source software; worst case its copyright expires and it falls into the public domain.

timoreimann: Done in #272.

PR #262, opened 2020-01-08
Enable block volumes in upstream e2e tests

timoreimann:
With support for raw block volumes available, we should enable the corresponding upstream end-to-end tests.


timoreimann:
Upstream storage end-to-end tests passed.

PR #266, opened 2020-01-10
Update csi-snapshotter to v1.2.2

timoreimann:
Also update RBAC rules and CRD definitions (which got amended by a subresources field).

I initially rejected to update csi-snapshotter in #236 because of a breaking CRD change; on closer inspection, it turned out that the delta is minimal (a new subresource was added), so we should include this one.


timoreimann:
Upstream storage end-to-end tests passed.

PR #268, opened 2020-01-11
Fix Travis configuration to call right Makefile target

timoreimann:
Our "all" target is supposed to be suitable for CI, but we did not use it.


timoreimann:
The PR is failing because of #269 (which also shows that this PR is working).

timoreimann:
Rebased; Travis is green now as expected.

PR #269, opened 2020-01-11
Update Go modules files

timoreimann:
`go mod {tidy,vendor}` was not invoked after the last dependency update.


timoreimann:
The referenced PR fixes our Travis CI configuration to make sure this won't happen again.

PR #274, opened 2020-01-14
Implement CI using Github Actions

timoreimann:
**tl;dr:** Test execution (unit / end-to-end) and releasing is now handled via Github Actions

This change implements a continuous integration (CI) workflow using Github Actions.

In particular, we configure workflows to run unit and end-to-end tests on every PR (update) and push to the master branch. This includes the necessary steps to build test runner and plugin images and execute upstream storage tests against a dynamically created DOKS cluster.

We also trigger the release process when a tag is pushed, including the publication of an official Docker Hub container image as well as the creation of a Github release.

This change obsoletes our usage of Travis CI.

See the updated markdown files for usage instructions.

Other release branches would probably benefit from a similar change. I will look into those after this PR gets merged.


timoreimann:
I tested this change on my fork. To see things in action, see

- the [PR test run](https://github.com/timoreimann/csi-digitalocean/commit/32b4e50f94de807c00ccbd608cc6ef0935780c90/checks?check_suite_id=399390844) (pre-merge)
- the [master branch test run](https://github.com/timoreimann/csi-digitalocean/commit/287a6e6726764bb6199aefb85212da6e0aca23fc/checks?check_suite_id=399586613) (post-merge)
- the [release run](https://github.com/timoreimann/csi-digitalocean/commit/c671c361cfcc4ebad4f97bdba7d0108fba3db6a0/checks?check_suite_id=399766403) (on tag push)

Chances are this PR may need further adjustments to finish the transition from my repo to this one; we'll see as the workflows progress in Github Actions.

test/e2e/testdrivers/1.16.yaml
```diff
@@ -25,8 +25,8 @@ DriverInfo:
     exec: true
     snapshotDataSource: true
     multipods: true
-    controllerExpansion: true
-    nodeExpansion: true
+    controllerExpansion: false
```
timoreimann: Disabled while https://status.digitalocean.com/incidents/qz40dp1qcs1d is still ongoing; I won't merge the PR before undoing this file change and seeing a green run.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+  e2e-test:
+    runs-on: ubuntu-18.04
+    needs: push-images
+    strategy:
+      matrix:
+        kube-release: ['1.16', '1.15', '1.14']
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
```
nicktate: ditto previous comments in release workflow rg. actions/checkout

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+  pull_request:
+    branches:
+
+jobs:
+  unit-test:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
```
nicktate: ditto previous comments in release workflow rg. actions/checkout

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+        kube-release: ['1.16', '1.15', '1.14']
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Go setup
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.13.5'
```
nicktate: nit: lets make this a top level env var so it is more prominently located if we need to change

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+    needs: unit-test
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Log into container registry
+        run: echo ${{ secrets.DockerHubToken }} | docker login --username ${DOCKER_USER} --password-stdin
+
+      - name: build and push runner image
```
nicktate: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths

Maybe we can limit when this actually triggers with the `path` option? Although I see you've done some smart `--from-cache` work with the build so it may not be needed

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+        env:
+          # ${{ github.head_ref }} is empty for non-pull requests, which is
+          # exactly what we want on master.
+          RUNNER_IMAGE_TAG_PREFIX: ${{ github.head_ref }}
```
nicktate: It doesn't seem like we are using this env var in the run command for this step

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+        run: RUNNER_IMAGE=${DOCKER_ORG}/k8s-e2e-test-runner make runner-push
+
+      - name: Build and push plugin image
+        env:
+          VERSION: ${{ github.head_ref }}
+        run: |
+          VERSION="${VERSION:-latest}"
```
nicktate: Should we be defaulting to `master` ?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Go setup
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.13.5'
+
+      - name: Run unit tests
+        run: make
```
nicktate: I think it might be safer to explicitly specify the make command in case default functionality would be updated in the Makefile for whatever reason

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+    runs-on: ubuntu-18.04
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Go setup
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.13.5'
```
nicktate: nit: lets make this a top level env var so it is more prominently located if we need to change

.github/workflows/test.yaml
```diff
@@ -0,0 +1,131 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Log into container registry
+        run: echo ${{ secrets.DockerHubToken }} | docker login --username ${DOCKER_USER} --password-stdin
+
+      - name: tag new CSI plugin master image
+        run: |
+          docker pull ${DOCKER_ORG}/do-csi-plugin-dev:latest
```
nicktate: Do we need latest if we just build as `master`? I guess I could see an argument in terms of semantics but they will always be in-sync as far I can tell?

Makefile
```diff
@@ -86,6 +97,160 @@ endif
 	@docker push $(DOCKER_REPO):$(VERSION)
 	@echo "==> Your image is now available at $(DOCKER_REPO):$(VERSION)"
 
+# runner-build builds the e2e test runner image. Sadly, this is much harder to
+# do cache-efficiently than it should be for Docker multi-stage builds since
+# those require build targets to be pushed and referenced individually. See
+# https://andrewlock.net/caching-docker-layers-on-serverless-build-hosts-with-multi-stage-builds---target,-and---cache-from/
```
nicktate: 😞 that this needs to be done but 👏 amazing job discovering a way to utilize cache to its max

.github/workflows/test.yaml
timoreimann: I tried this; unfortunately, environment variables are not expandable here. :(

.github/workflows/test.yaml
timoreimann: Unfortunately, the path-based trigger does not work at the job or step level, and there's no way to describe dependencies between workflows, so we cannot use it.

.github/workflows/test.yaml
timoreimann: I chose `latest` because it makes the usage of the image slightly easier (you do not need to specify that tag) and it minimizes risk for confusion with the plugin image.

The reason I went with `master` for the plugin image `digitalocean/do-csi-plugin` is that I did not want to give a false impression to users that the latest tag is an alternative name for whatever the latest _stable_ release is (e.g., v1.2.0 as of now), which is usually how `latest` is used.

.github/workflows/test.yaml
timoreimann: Oh yeah, great call. 👍

.github/workflows/test.yaml
timoreimann: My take on this is that `master` is the final product of our CI run that is known to have passed all tests successfully; other than for this final step in our CI, it is not meant to be otherwise updated but consumed only.

`latest` on the other hand is WIP until we have successfully completed the final step in our CI.  It's an implementation detail that could contain broken code until CI has finally assured correctness.

To put it in more practical terms: if we used a single `master` tag only, we could end up in a situation where we are updating our single image tag in the _Build and push plugin image_ step (as the PR does today) only to find it broken in the `Run end-to-end tests` step; at that point, our single `master` image broke, and so will potential (external) consumers.

Does my train of thought make sense?

Makefile
timoreimann: Looking for the German word that expresses feeling both pride and rejection about one's work. :)

timoreimann:
Thanks for the super useful feedback @nicktate. There were a few I responded to, but most of them I was happy to accept right away.

I pushed another commit; I'm going to check if it works as expected and if so, do another round of testing in my personal repo for the release part.

.github/workflows/test.yaml
nicktate: Ahh, I see what you are saying, that makes sense to me. I guess the only thing that comes to mind as a minor concern is if we have two PRs merged near one another, then the builds will be competing for contents of that tag.

I'm not sure if it makes sense to use a commit hash based approach for this intermediate build and then cleanup later on versus clashing on master. I will leave that up to you, I think for the most part we shouldn't run into issues.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: Log into container registry
+        run: echo ${{ secrets.DockerHubToken }} | docker login --username ${DOCKER_USER} --password-stdin
```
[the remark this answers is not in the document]
nicktate: I think there were two instances of the double quote, it looks like you missed this one and fixed the latter.

.github/workflows/test.yaml
timoreimann: Hmm yeah that's a fair point. What I'm not sure is if two consecutive master builds would be able to compete. Does the latter perhaps cancel the former build? I haven't really tested this to be honest.

I'm inclined to giving this a try as it stands now and adjust if it turns out to be problematic.

.github/workflows/release.yaml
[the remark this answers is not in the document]
timoreimann: Oh you're right, that somehow slipped through. I should have searched-and-replaced in my editor.

Thanks, going to fix this.

timoreimann:
Full test and release cycle was found to be working in my personal repo. I'm going to move forward and merge this one.

Thanks again for the review @nicktate -- I like how we were able to improve the implementation.

PR #280, opened 2020-01-23
e2e tests: keep cluster and node pool names at valid lengths

timoreimann:
Our existing cluster and node pool naming schema may include components of indeterministic length, such as the branch name in CI. However, DOKS rejects cluster and node pool names that are too long to ensure conformity with DNS naming rules imposed by Kubernetes.

This change makes sure that cluster and node pool names stay within valid bounds. Specifically, we limit the cluster name by including the shorter Kubernetes major/minor version only. Additionally, we
name-suffix cluster names with the hash of the branch name instead of the branch itself during CI. (We do not do this in general since interactive users likely want readable names and should know what they are doing.)

Based on #278 since we leverage the major/minor version extraction which that PR introduced.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -111,6 +111,9 @@ jobs:
           NAME_SUFFIX="${BRANCH}"
           TAG="${BRANCH}"
           if [[ $BRANCH != "" ]]; then
+            # Hash name suffix which goes into the cluster name to ensure we do
+            # not exceed any name constraints.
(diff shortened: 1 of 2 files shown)


timoreimann:
@nicktate I considered trimming, but there's at least a theoretical risk that branches `fix-bug-in-controller` and `fix-bug-in-node` lead to the same common suffix if the trimming is on too few characters. That's why I went with a hash in CI (not very readable, but practically eliminates the possibility for equally named clusters) and put it on the user to pick proper suffix names if the end-to-end tests are executed manually.

PR #281, opened 2020-01-26
Improve debug HTTP server usage

timoreimann:
- Rename the `-address` parameter to `-debug-addr`.
- Only start the HTTP server on the Controller service: the Node service does not have a token configured and thus would never be able to respond successfully to a request.
- Only start the debug server if `-debug-addr` is given, ensuring that existing users won't be surprised by the new functionality. This also allows us to test a dev version of the CSI plugin on a cluster where  the port might already be used.
- Remove explicit HTTP listener which is encapsulated by HTTP server.
- Bound graceful HTTP server shutdown by 10 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## unreleased
 
+* Improve debug HTTP server usage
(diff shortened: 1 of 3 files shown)


timoreimann:
This change is needed to unblock other pending PRs.

PR #284, opened 2020-02-07
Update alpine to 3.11

varshavaradarajan:


varshavaradarajan:
Closing, since the branch name causes an invalid reference format while tagging the docker image. Will fix that in another PR.

PR #285, opened 2020-02-07
Update alpine to 3.11

varshavaradarajan:
Update base image for the csi plugin to 3.11


timoreimann:
Hmm, it somehow looks like we have status check results from multiple runs (with [the latest one actually succeeding](https://github.com/digitalocean/csi-digitalocean/actions/runs/35776754)). I'm going to retrigger another build, hoping that things will clean up.

PR #287, opened 2020-02-17
delete PR images from Dockerhub after merge

varshavaradarajan:
When a PR is either closed or merged, this job will run to delete the associated images from Dockerhub repos: https://hub.docker.com/r/digitalocean/do-csi-plugin-dev and https://hub.docker.com/r/digitalocean/k8s-e2e-test-runner. 

If the image is not available, then the API requests succeeds with 200 but response just has `Not found`. We ignore and continue in such cases.

### DIFF
--- a/.github/workflows/delete.yaml
+++ b/.github/workflows/delete.yaml
@@ -1,0 +1,20 @@
+name: delete-pr-images
+on:
+  pull_request:
+    types: [closed]
+jobs:
+  delete_tags:
+    name: Delete PR images from Dockerhub
+    runs-on: ubuntu-latest
+    steps:
+      - name: dockerhub task
+        env:
+          BRANCH: ${{ github.head_ref }}
+          DOCKER_USER: ${{ secrets.DockerHubUser }}
+          DOCKER_PASS: ${{ secrets.DockerHubToken }}
+        run: |
+          BRANCH=$(echo -n ${BRANCH} | tr -c '[:alnum:]._-' '-')
+          TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username": "'${DOCKER_USER}'", "password": "'${DOCKER_PASS}'"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
+          images=("${BRANCH}-latest" "${BRANCH}-runtime" "${BRANCH}-tools" "${BRANCH}-tests-1.14" "${BRANCH}-tests-1.15" "${BRANCH}-tests-1.16" "${BRANCH}-builder" "${BRANCH}-builder-pre-1.16")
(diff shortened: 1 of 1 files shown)


adamwg:
Don't want to derail this, since it looks like you've got a solution that works, but a handy tool for working with container registries is [reg](https://github.com/genuinetools/reg). In particular, it can list and delete tags (and also delete images by digest, which may be desirable here). It uses the credentials persisted by `docker login` and will do the JWT auth dance when necessary.

varshavaradarajan:
@timoreimann - if the job to build image fails and there is no image to delete when the PR gets closed, the API request to delete the image succeeds with 200 but shows `{"detail": "Not found"}`. But, I think we should add `|| true` in case the request fails for auth or too many requests etc. Thanks! :) 

@adamwg - thanks! I did take a look at `reg` (you had mentioned it on a team call, I think). But because there were install steps involved, I chose to use curl and jq which is available right away. But, why do you think that deleting an image by the name isn't enough and we might want to use a digest?

adamwg:
@varshavaradarajan When you delete a tag, the image is still available via its digest (similarly, if you push a new version of an image using the same tag, you can still pull the old version by digest). When you delete a tag, the tag reference gets deleted but the digest reference and underlying layer data does not.

varshavaradarajan:
@adamwg - thanks! I didn't know that. But, deleting by digest requires the environment variable `REGISTRY_STORAGE_DELETE_ENABLED ` to be set. And Dockerhub seems to have it disabled, so does not allow delete by manifest. I tried to use the registry API https://docs.docker.com/registry/spec/api/#deleting-an-image, I get `405 Not allowed`. 

This was my request:

```
curl -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/<image-name>/manifests/sha256:xxx
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>openresty</center>
</body>
</html>
```

Found https://github.com/docker/hub-feedback/issues/1759#issuecomment-470317472 and https://github.com/davedoesdev/dxf/issues/25#issuecomment-533912742 finally. I  am not sure if these unreferenced manifests get pruned by DockerHub later on.

adamwg:
@varshavaradarajan Interesting - thanks for exploring that. There's definitely value in cleaning up the tags regardless, so no objection to merging this as-is. :+1:

varshavaradarajan:
@timoreimann - some more unexpected  DockerHub behaviour. Delete doesn't happen because of `{"detail": "access to the resource is forbidden with personal access token"}`. I checked lastpass and there doesn't seem to be a password there. Can you please update lastpass if you know what it is? Thanks!

PR #294, opened 2020-03-27
Add example describing how to import a snapshot

### LINKED ISSUE #292: Is it possible to create a VolumeSnapshot in kubernetes from an existing snapshot in digitalocean?
[pod-single-existing-volume](https://github.com/digitalocean/csi-digitalocean/tree/master/examples/kubernetes/pod-single-existing-volume) shows how to add an existing volume as a PersistentVolume by setting an annotation:
  
    annotations:
      # fake it by indicating this is provisioned dynamically, so the system
      # works properly
      pv.kubernetes.io/provisioned-by: dobs.csi.digitalocean.com

But I haven't found a way to do the same for an existing snapshot. Is it currently possible to do that?

timoreimann:
Fixes #292

### DIFF
--- a/examples/kubernetes/import-snapshot/README.md
+++ b/examples/kubernetes/import-snapshot/README.md
@@ -21,7 +21,7 @@ spec:
   csiVolumeSnapshotSource:
     driver: dobs.csi.digitalocean.com
     snapshotHandle: 92b46522-4798-4544-9d33-8b74ea56adb7
-  deletionPolicy: Delete
+  deletionPolicy: Retain
   volumeSnapshotRef:
     apiVersion: snapshot.storage.k8s.io/v1alpha1
     kind: VolumeSnapshot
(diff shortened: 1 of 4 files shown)


examples/kubernetes/import-snapshot/README.md
```diff
@@ -0,0 +1,91 @@
...
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1alpha1
+kind: VolumeSnapshotContent
+metadata:
+  name: snapshotcontent-manual
+spec:
+  csiVolumeSnapshotSource:
+    driver: dobs.csi.digitalocean.com
+    snapshotHandle: 92b46522-4798-4544-9d33-8b74ea56adb7
+  deletionPolicy: Delete
```
adamwg, line 24: I wonder if we should recommend `deletionPolicy: Retain` as the default here. I'm sure there are cases where users want the cluster to "own" the snapshot and delete it when they're done with it, but since it was created out-of-band it seems reasonable that many users will also want to delete it out-of-band.

timoreimann: Makes total sense to me. I flipped the field to `Retain` and explained the implications of the two field options below.

PR #299, opened 2020-04-17
Support filtering snapshots by IDs

timoreimann:
Our implementation of `ListSnapshots` currently returns all snapshots. The request parameter also provides fields to filter by snapshot ID and source volume ID, which we have been ignoring so far.

Based on [discussions with SIG Storage](https://kubernetes.slack.com/archives/C8EJ01Z46/p1587038433091900), it became clear that CSI
drivers are expected to process present IDs in the `ListSnapshots` call appropriately. While the spec marks them as OPTIONAL, it turned out that the optionality refers to the CO, not the SP. (A request for
clarification [has been filed](https://github.com/container-storage-interface/spec/issues/426).) As soon as any of IDs are provided, drivers are expected to take them into account.  In fact, csi-snapshotter always passes a snapshot ID in and only ever takes the first snapshot item from the list of snapshots returned, which means that our current implementation is buggy for cases where the account
holds more than one snapshot (at least on Kubernetes; other COs could potentially behave differently).

This change fixes the issue by retrieving the snapshot by direct lookup if a snapshot ID is given.

If a source volume ID is given, we do additional filtering by only returning the snapshots that were sourced by the same volume.

The existing csi-sanity tests did not catch this bug. A [PR to that project](https://github.com/kubernetes-csi/csi-test/pull/259) was submitted to close the gap. I tested it locally and verified that the test now works. Once it is merged, we should upgrade our version of the test package.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -772,92 +772,118 @@ func (d *Driver) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotRequ
 // ListSnapshots shold not list a snapshot that is being created but has not
 // been cut successfully yet.
 func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsRequest) (*csi.ListSnapshotsResponse, error) {
-	// Pagination in the CSI world works different than at DO. CSI sends the
-	// `req.MaxEntries` to indicate how much snapshots it wants. The
-	// req.StartingToken is returned by us, if we somehow need to indicate that
-	// we couldn't fetch and need to fetch again. But it's NOT the page number.
-	// I.e: suppose CSI wants us to fetch 50 entries, we only fetch 30, we need to
-	// return NextToken as 31 (so req.StartingToken will be set to 31 when CSI
-	// calls us again), to indicate that we want to continue returning from the
-	// index 31 up to 50.
-
-	var nextToken int
-	var err error
-	if req.StartingToken != "" {
-		nextToken, err = strconv.Atoi(req.StartingToken)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


driver/controller.go
```diff
@@ -772,93 +772,119 @@ func (d *Driver) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotRequ
...
-		page, err := resp.Links.CurrentPage()
-		if err != nil {
-			return nil, err
+		// fetch all entries
+		listOpts := &godo.ListOptions{
+			PerPage: int(req.MaxEntries),
 		}
+		var snapshots []godo.Snapshot
+		for {
+			snaps, resp, err := d.snapshots.ListVolume(ctx, listOpts)
```
nicktate, line 835: I'm a little confused with this pagination section below. It seems like we paginate through every entry from the start, ignoring the offset of `NextToken` that is returned. Can we just determine the "pageNumber" in DO pagination terms as a calculation of the pageCount and the offset?

timoreimann: I think you're right, the logic looks off. I did not pay much attention to it because existing code was just moved into an `else` branch but was otherwise not modified. So I believe the pagination had been broken before. Great catch. 👍 

If you don't mind then I'd like to address this issue in a separate PR. FWIW, the flawed pagination should not have impacted users much since csi-snapshotter always passes in a snapshot ID in the `ListSnapshots` request, and our handling of that parameter is only fixed by this parameter. Being good CSI citizens, however, we should still fix the pagination for the case that `ListSnapshots` is ever consumed exhaustively either in the future or by other COs (container orchestrators) today already.

nicktate: :+1: That works for me.

timoreimann: Fix submitted in #300.

PR #300, opened 2020-04-20
Fix ListSnapshots paging

timoreimann:
This changes provides the following fixes and improvements to `ListSnapshots`:

- Use paging to collect snapshots beyond the first page. Previously, we would only return snapshots from the first page.
- Handle `StartingToken` and `MaxEntries` such that we use paging efficiently and skip initial, unneeded snapshots.
- Extend fake snapshot driver to support paging.
- Add tests.

Note that Kubernetes / the csi-snapshotter sidecar currently do not invoke `ListSnapshots` without the snapshot ID parameter, which means that the fixed code is not executed in production. However, it is used by csi-test / the sanity package, and other COs (Container Orchestrators) may potentially use it as well as Kubernetes going forward.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -776,12 +776,14 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	log := d.log.WithFields(logrus.Fields{
 		"snapshot_id":        req.SnapshotId,
 		"source_volume_id":   req.SourceVolumeId,
+		"max_entries":        req.MaxEntries,
 		"req_starting_token": req.StartingToken,
 		"method":             "list_snapshots",
 	})
 	log.Info("list snapshots is called")
 
 	if req.SnapshotId != "" {
+		// Fetch snapshot directly by ID.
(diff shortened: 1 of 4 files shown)


driver/controller.go
```diff
@@ -802,44 +804,97 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+			// hasMore indicates if NextToken must be set.
+			hasMore   bool
+			snapshots []godo.Snapshot
+		)
 		for {
+			hasMore = false
 			snaps, resp, err := d.snapshots.ListVolume(ctx, listOpts)
 			if err != nil {
-				return nil, status.Errorf(codes.Aborted, "ListSnapshots listing volume snapshots has failed: %s", err.Error())
+				return nil, status.Errorf(codes.Internal, "ListSnapshots listing volume snapshots has failed: %s", err)
```
nicktate, line 865: This status code was actually correct before: https://github.com/container-storage-interface/spec/blob/master/spec.md#listsnapshots-errors

PR #303, opened 2020-04-23
Enable tests for node volume attachment limits

timoreimann:
The csi-test sanity package ships with off-by-default [tests to validate per-node attachment limits](https://github.com/digitalocean/csi-test/blob/20e87c55f1f6f5694363f512bba52c3a38f5a6a1/pkg/sanity/controller.go#L1068). This change toggles the corresponding test configuration flag to enable the tests.

The change requires modifying our fake driver to return a 422 HTTP error code when the limit is exceeded. As a consequence, we also need to customize the IdempotentCount test setting which parameterizes the ['should be idempotent' test](https://github.com/digitalocean/csi-test/blob/20e87c55f1f6f5694363f512bba52c3a38f5a6a1/pkg/sanity/controller.go#L1288) that creates the given number of volumes in
sequence. The default value of 10 causes our (fake) limit to be exceeded, which is why we tune it down to 5.

The test also revealed that we missed to handle the case where the node volume attachment limit is exceed during `ControllerPublishVolume`. We extend our error handling to identify this case and return an `RESOURCE_EXHAUSTED` code accordingly.

The PR depends on #301.

### DIFF
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -96,6 +96,8 @@ func TestDriverSuite(t *testing.T) {
 
 	cfg := sanity.NewTestConfig()
 	cfg.Address = endpoint
+	cfg.IdempotentCount = 5
+	cfg.TestNodeVolumeAttachLimit = true
 	sanity.Test(t, cfg)
 
 	cancel()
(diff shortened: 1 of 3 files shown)


driver/driver_test.go
```diff
@@ -248,11 +250,72 @@ type fakeStorageActionsDriver struct {
...
+
+	droplet, ok := f.droplets[dropletID]
+	if !ok {
+		resp.Response = &http.Response{
+			StatusCode: http.StatusNotFound,
+		}
+		return nil, resp, errors.New("droplet was not found")
+	}
+
+	if len(droplet.VolumeIDs) == maxVolumesPerNode {
```
adamwg, line 270: Should we use `>= maxVolumesPerNode` here just to be safe?

timoreimann: Yep, good call -- done.

PR #307, opened 2020-04-29
Upgrade Kubernetes dependencies to 1.17

timoreimann:
Details:

- Add script and Makefile target to easily update Kubernetes dependencies.
- Update Kubernetes dependencies to 1.17.5.
- Adjust code to dependency refactorings.


scripts/update-k8s.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e
+
+KUBERNETES_VERSION=${KUBERNETES_VERSION:?}
+
+deps=()
+
+while read -ra LINE
```
waynr, line 8: I took your suggestion from slack the other day: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/update-k8s.sh

So instead of the dependencies that are already listed in the `go.mod` file the script linked above curls the upstream kubernetes `go.mod` at the appropriate version. It also moves the `k8s.io/kubernetes` dependency outside the loop since of course the upstream `go.mod` doesn't list itself as a dependency the same way it lists its components as dependencies and we do still need a `replace` directive for it.

timoreimann: I did pull down the CCM version of that script but then failed by committing an old, local copy I still had. :)

Should be fixed now.

PR #308, opened 2020-04-29
Run 1.17 upstream E2E tests

timoreimann:
This change adds the Kubernetes 1.17 upstream E2E tests to our CI, validating that our driver is compatible with that Kubernetes release.

One caveat we have is that the Kubernetes 1.17 E2E tests expect the snapshot API to be used in the beta version whereas our driver still uses alpha. We work around this issue by excluding the snapshot-related tests from the 1.17 suite and run run the 1.16 snapshot-only tests against a 1.17 cluster instead. Once we have migrated over to the beta API, we can run the full 1.17 suite again and benefit from the latest snapshot tests in upstream.

The deviation requires making a few adjustments to our E2E test in terms of allowing more flexible combinations of testdriver configuration files and target Kubernetes releases to run the tests against. Similarly, our Github Actions specification is amended to reflect the new CLI flags added to our E2E test, and we add another job for the 1.16 snapshot-only tests leveraging [Github Actions combinations](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations).

Building our runner image for 4 Kubernetes now drives us into a [disk space limitation set on the Github Actions workers](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources), which is why we resort to cleaning up unneeded files from the workers before the start of the build process. (Eventually, we should find a better way to deal with this constraint, perhaps through splitting up our build steps.)

Finally, we include a few improvements to our cluster setup scaffolding to achieve better error reporting and resilience towards transient errors.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -59,16 +59,30 @@ Command-line arguments are passed as-in to the test tool. Run `e2e.sh -h` for us
 ### Run the end-to-end tests
 
 1. If necessary, update the test runner image (see below for instructions)
-2. If necessary, add or update the testdriver file definitions (see the [related section above](#testdrivers) for details)
-3. Execute `e2e.sh`, passing in parameters as needed
+1. If necessary, add or update the testdriver file definitions (see the [related section above](#testdrivers) for details)
+1. Execute `e2e.sh`, passing in parameters as needed
 
-### Update the test runner image
+### Update an existing test runner image
 
 1. Make and push any necessary updates to our Kubernetes fork
-2. If the fork was updated: Update the `SHA_*` build argument(s) in the Dockerfile to point to the new commit hash(es)
-3. Run `handle-image.sh build`
-4. Run `handle-image.sh push`
+1. If the fork was updated: Update the `SHA_*` build argument(s) in the runner image Dockerfile to point to the new commit hash(es)
+1. Add the build argument
 
-For testing purposes, you may also overwrite a commit hash on-the-fly by defining the corresponding environment variable(s) `SHA_*`, e.g.: `SHA_1_16=deadbeef ./handle-image.sh build`.
+### Add support for a new Kubernetes release
+
+1. Make and push any necessary updates to our Kubernetes fork
+1. Add a new Kubernetes version-specific block to the runner image Dockerfile; make sure to update the `SHA_*` variable as well
+1. Update the kubectl version in the test runner Dockerfile
+1. Update the Makefile `runner-build` and `runner-push` targets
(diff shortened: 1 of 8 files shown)


timoreimann:
The CI run was green before, so the PR should be good for review.

test/e2e/README.md
```diff
@@ -59,16 +59,30 @@ Command-line arguments are passed as-in to the test tool. Run `e2e.sh -h` for us
...
 
 1. Make and push any necessary updates to our Kubernetes fork
-2. If the fork was updated: Update the `SHA_*` build argument(s) in the Dockerfile to point to the new commit hash(es)
-3. Run `handle-image.sh build`
-4. Run `handle-image.sh push`
+1. If the fork was updated: Update the `SHA_*` build argument(s) in the runner image Dockerfile to point to the new commit hash(es)
+1. Add the build argument
 
-For testing purposes, you may also overwrite a commit hash on-the-fly by defining the corresponding environment variable(s) `SHA_*`, e.g.: `SHA_1_16=deadbeef ./handle-image.sh build`.
+### Add support for a new Kubernetes release
```
waynr, line 71: :100:

test/e2e/e2e_test.go
```diff
@@ -335,21 +354,26 @@ func createCluster(ctx context.Context, client *godo.Client, nameSuffix, kubeMaj
 		page = p + 1
 	}
 
+	if deleted {
+		// Sleeping briefly to reach consistent state.
```
waynr, line 358: I am assuming by "reach consistent state" you mean "don't try to create before the old cluster has finished deleting"; what do you think about a wait loop that waits until a `client.Kubernetes.Get` (or whatever the equivalent method is called) returns a 404?

Probably not a huge deal since we don't run this suite very often, but we do do something like that in our internal test runner suite.

timoreimann: That makes a lot of sense to me. I implemented your suggestion and updated the PR accordingly.

While I was doing so, I realized why the PR would fail frequently. Initially I blamed it on Github [Actions], but then it turned out that the extra snapshot-only E2E test run I added was using the same cluster name as the 1.17 run. Consequently, they were racing for the same cluster, which is why one might see a 404 right after the cluster create when it overlapped with the other run deleting what it thought to be a left-over cluster from a previous run.
I made a small modification to the workflow to add another suffix to the cluster name when we run the snapshot-only tests. It's not pretty but it gets the job done until we can (hopefully soon) use the full 1.17 E2E suite from upstream.

PR #310, opened 2020-05-01
Fix ListVolumes paging

timoreimann:
Previously, the `ListVolumes` implementation would ignore the starting token and max entries value from the request parameter, thereby effectively not using paging. This change provides a fix by refactoring the existing implementation for `ListSnapshots` into a reusable helper method and using it from both `List*` implementations.

#300 introduced the fixed paging method for `ListSnapshots`.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -520,67 +520,45 @@ func (d *Driver) ValidateVolumeCapabilities(ctx context.Context, req *csi.Valida
 
 // ListVolumes returns a list of all requested volumes
 func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (*csi.ListVolumesResponse, error) {
-	var page int
-	var err error
-	if req.StartingToken != "" {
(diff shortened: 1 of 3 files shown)


driver/controller.go
```diff
@@ -520,67 +520,45 @@ func (d *Driver) ValidateVolumeCapabilities(ctx context.Context, req *csi.Valida
...
+		volumes, resp, err := d.storage.ListVolumes(ctx, volListOpts)
+		if err != nil {
+			return nil, resp, err
 		}
 
-		page, err := resp.Links.CurrentPage()
-		if err != nil {
-			return nil, err
+		untypedVolumes := make([]interface{}, 0, len(volumes))
+		for _, snap := range volumes {
```
waynr, line 550: I am a little confused by the use of `snap` here, it looks like we later cast to a `godo.Volume`. I'm guessing it's just a copy pasta error.

timoreimann: Ah yes indeed -- good catch, it's fixed now.

PR #313, opened 2020-05-04
e2e: propagate container run exit code upwards

timoreimann:
This fixes a bug where a container run returning with a non-zero exit code would not be propagated upwards properly, thus making actually failed runs pass.

As a consequence, we also have to toggle a few flags for the 1.16 snapshot-only testdriver configuration to make the E2E tests actually pass.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -473,7 +473,7 @@ func deployDriver(ctx context.Context, driverImage string, kubeconfigFile, token
 // runE2ETests invokes our test container.
 // It passes in bind-mount parameters for the kubeconfig and the location of the
 // testdriver YAML files.
-func runE2ETests(ctx context.Context, kubeVersion, runnerImage, testdriverFilename, focus, kubeconfigFile, token string, skipParallel, skipSequential bool, ginkgoNodes int) (retErr error) {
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -473,7 +473,7 @@ func deployDriver(ctx context.Context, driverImage string, kubeconfigFile, token
 // runE2ETests invokes our test container.
 // It passes in bind-mount parameters for the kubeconfig and the location of the
 // testdriver YAML files.
-func runE2ETests(ctx context.Context, kubeVersion, runnerImage, testdriverFilename, focus, kubeconfigFile, token string, skipParallel, skipSequential bool, ginkgoNodes int) (retErr error) {
+func runE2ETests(ctx context.Context, kubeVersion, runnerImage, testdriverFilename, focus, kubeconfigFile, token string, skipParallel, skipSequential bool, ginkgoNodes int) error {
```
timoreimann, line 476: The named returned parameter was never used, so we turn it into an unnamed one.

timoreimann:
This is working as intended: [one of the jobs](https://github.com/digitalocean/csi-digitalocean/pull/313/checks?check_run_id=644054941) that needs #312 is failing right now.

Next up, I'll rebase with the fix included and (hopefully) watch the build go green.

PR #318, opened 2020-06-06
Update Kubernetes dependencies to 1.18.3

timoreimann:


go.mod
```diff
@@ -18,37 +18,37 @@ require (
...
-	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.17.5
-	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.17.5
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.17.5
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.17.5
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.17.5
-	k8s.io/kubectl => k8s.io/kubectl v0.17.5
-	k8s.io/kubelet => k8s.io/kubelet v0.17.5
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.17.5
-	k8s.io/metrics => k8s.io/metrics v0.17.5
+	k8s.io/api => k8s.io/api v0.18.3
```
adamwg, line 30: Is this `replace` necessary, given that `k8s.io/api v0.18.3` is in the `require` above?

go.mod
```diff
@@ -18,37 +18,37 @@ require (
...
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.18.3
+	k8s.io/apimachinery => k8s.io/apimachinery v0.18.4-rc.0
```
adamwg, line 32: Is this intentionally not `0.18.3`?

go.mod
timoreimann: It should not be, and a few other direct dependencies of ours probably don't need the replacement either. I think it's a consequence of how we upgrade dependencies through [our script](https://github.com/digitalocean/csi-digitalocean/blob/275ffede814edfa5ffa722a8ef1adc6a61d02217/scripts/update-k8s.sh), which we need because we have to vendor k8s.io/kubernetes for a resizing library, and that requires replacing all of the monorepo's dependencies in the fashion we do.

We could exclude the direct dependencies from the script, but it'd involve a bit effort for I think relatively little gain. AFAIK, upstream is planning on moving the resize logic into a separate repository as well; at that point, we can simplify our dependency management process by referencing direct dependencies only and letting gomod pull in just the transitive dependencies we need.

go.mod
timoreimann: This one's a bit odd: our upgrade script puts in a replacement rule to v0.18.3 properly. It's only when `go mod tidy` runs that k8s.io/apimachinery is replaced for v0.18.4-rc.0.

I played around with `go mod why` and `go mod graph` but couldn't find any indicators for why the release candidate would be referenced. I wonder if this is a bug. 🤔 

On the bright side, changing the replacement to v0.18.3 and running `go mod tidy` afterwards seems to work in the sense that the rc version is _not_ replaced again. That'd work for me for now unless you may have another idea on what we can or should do here.

timoreimann: I think I understand now:  k8s.io/apimachinery@v0.18.3 and k8s.io/apimachinery@v0.18.4-rc.0 point at the same commit. It seems similar with a few other dependencies.

I forced all of them into v0.18.3, it makes sense to me to be consistent.

PR #319, opened 2020-06-06
Replace irritating usage of sed with in-place edits

timoreimann:
~One of sed's peculiarities is that it leaves behind backup files with a `-e` suffix. Unfortunately, it is not easy to skip those in an operating system-agnostic way.~

Our (portable) usage of `sed -i` is rather irritating because it gives the impression that we use sed's `-e` parameter when instead this becomes our backup suffix. We replace this kind of usage by a more straight-forward syntax.

See [my comment below](https://github.com/digitalocean/csi-digitalocean/pull/319#discussion_r436249610) on the differences between BSD and GNU sed, and why there's no way to omit backup files in a portable fashion.

### DIFF
--- a/scripts/update-k8s.sh
+++ b/scripts/update-k8s.sh
@@ -39,5 +39,5 @@ go mod tidy
 go mod vendor
 set +x
 
-sed -i -e "s/^KUBERNETES_VERSION.*/KUBERNETES_VERSION ?= $KUBERNETES_VERSION/" Makefile
-rm Makefile-e
+sed -i.sedbak "s/^KUBERNETES_VERSION.*/KUBERNETES_VERSION ?= $KUBERNETES_VERSION/" Makefile
+rm -f Makefile.sedbak
--- a/scripts/update-k8s.sh
+++ b/scripts/update-k8s.sh
@@ -40,3 +40,4 @@ go mod vendor
 set +x
 
 sed -i -e "s/^KUBERNETES_VERSION.*/KUBERNETES_VERSION ?= $KUBERNETES_VERSION/" Makefile
+rm Makefile-e
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 cmd/do-csi-plugin/do-csi-plugin
 .idea/
+*.sedbak
--- a/Makefile
+++ b/Makefile
@@ -57,13 +57,13 @@ bump-version:
 	@echo "Bumping VERSION from $(VERSION) to $(NEW_VERSION)"
 	@echo $(NEW_VERSION) > VERSION
 	@cp deploy/kubernetes/releases/csi-digitalocean-latest.yaml deploy/kubernetes/releases/csi-digitalocean-${NEW_VERSION}.yaml
-	@sed -i'' -e 's#digitalocean/do-csi-plugin:dev#digitalocean/do-csi-plugin:${NEW_VERSION}#g' deploy/kubernetes/releases/csi-digitalocean-${NEW_VERSION}.yaml
+	@sed -i.sedbak 's#digitalocean/do-csi-plugin:dev#digitalocean/do-csi-plugin:${NEW_VERSION}#g' deploy/kubernetes/releases/csi-digitalocean-${NEW_VERSION}.yaml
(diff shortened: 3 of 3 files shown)


scripts/update-k8s.sh
```diff
@@ -40,3 +40,4 @@ go mod vendor
 set +x
 
 sed -i -e "s/^KUBERNETES_VERSION.*/KUBERNETES_VERSION ?= $KUBERNETES_VERSION/" Makefile
```
waynr, line 42: Actually the problem here is the `-e`:

```
       -i[SUFFIX], --in-place[=SUFFIX]

              edit files in place (makes backup if SUFFIX supplied)
```

On my machine the suffix is only included when there is no space between the `-i` the `-e` which makes me think there must be a difference in the version of `sed` you are running. Regardless of why, it turns out the `-e` isn't necessary with my sed so I would try removing that before we add the `rm Makefile-e`. (also: all dev scripts like this should run in a devtools container in my opinion :wink:, not that i think we need to do anything that drastic here and now)

```
zsh/4 10018  (git)-[master]-% sed -h
sed: invalid option -- 'h'
Usage: sed [OPTION]... {script-only-if-no-other-script} [input-file]...
```

Also as a tangential aside, I love the fact that `-h` is an invalid option yet passing it gave me the intended usage string anyway.

timoreimann: The problem is in the difference between BSD and GNU sed, and it's actually about the `-i` parameter.

Here's the relevant part of what the BSD man-page has to say about that parameter:

```
     -i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no backup will be saved.
```

Note that a zero-length extension must be given to ignore backups.

In comparison, this is the section from the GNU man-page:

```
-i[SUFFIX], --in-place[=SUFFIX]
edit files in place (makes backup if extension supplied).
```

Note that this version does not expect (nor support) a zero-length parameter.

The difference means that on Linux, you can do

```shell
$ echo foo > foo
$ sed -i 's/foo/bar/' foo
$ cat foo
bar
$ ls
foo
```

but you cannot do

```shell
$ sed -i '' 's/foo/bar/' foo
sed: can't read s/foo/bar/: No such file or directory
```

because no parameter is expected; though interestingly, when you omit the space between `-s` and `''` it works

```shell
$ echo foo > foo
$ sed -i'' 's/foo/bar/' foo
$ cat foo
bar
$ ls
foo
```

(presumably because you're concatenating `-i` with the empty string, yielding `-i` again).

Now on a Mac, you cannot do

```shell
$ echo foo > foo
$ sed -i 's/foo/bar/' foo
sed: 1: "foo": invalid command code f
```

because BSD sed expects a parameter. We can give it the zero-length extension and it'll work fine:

```shell
$ echo foo > foo
$ rm -f foo-e
$ echo foo > foo
$ sed -i '' 's/foo/bar/' foo
$ cat foo
bar
$ ls
foo
```

Unlike GNU sed, however, it needs the space between `-s` and `''`:

```shell
$ echo foo > foo
$ sed -i'' 's/foo/bar/' foo
sed: 1: "foo": invalid command code f
```

As I realize now, the the way we've been including the `-e` parameter is actually quite misleading: when you do this on a Mac

```shell
$ echo foo > foo
sed -i'' -e 's/foo/bar/' foo
$ cat foo
bar
$ ls
foo  foo-e
```

what you've actually done is specify `-e` as the backup file extension parameter (with the `''` part again not being relevant due to the missing space). And you get the exact same result when you omit the `''`. In particular, we're not using the `-e` sed parameter which again seems to have slightly different semantics on BSD sed

```
   -e command
             Append the editing commands specified by the command argument to the list of commands.

```

vs. GNU sed:

```
-e script, --expression=script
add the script to the commands to be executed
```

This means two things:

1. There is no portable way to omit the creation of the backup file. The only portable way to do in-place replacements with sed is to create the backup file and delete it afterwards.
1. The `-e` usage as we've done it has been pretty irritating.

I agree with you that we should have a tools container, or alternatively expect Mac users of this repo to install BSD sed via brew. The latter is somewhat inconvenient for our casual external contributors, however.

This concludes my StackOverflow-ish response. For now I'm going to change out sed usage everywhere to at least make it more obvious what we're doing. Happy to hear your opinion on the PR update I'm going to make in a bit.

Issue #328, opened 2020-06-16
csi-do-controller-0 CrashLoopBackOff


max3903:
### What did you do? (required. The issue will be **closed** when not provided.)

I followed the documentation to add the do-block-storage plugin:

I added the secret successfully and run:

```
kubectl apply -f https://raw.githubusercontent.com/digitalocean/csi-digitalocean/master/deploy/kubernetes/releases/csi-digitalocean-v1.3.0.yaml
```

It fails on some snapshot specific stuff:
```
CustomResourceDefinition.apiextensions.k8s.io "volumesnapshots.snapshot.storage.k8s.io" is invalid: spec.version: Invalid value: "v1alpha1": must match the first version in spec.versions
```

I moved on and tried to create a PVC.

### What did you expect to happen?

I was expecting the PV to be created.

### Configuration (**MUST** fill this out):

* system logs:

https://gist.github.com/max3903/acb18527be1138a33d77f3eaaddb89a8

* manifests, such as pvc, deployments, etc.. you used to reproduce:

secret.yaml:

```
apiVersion: v1
kind: Secret
metadata:
  name: digitalocean
  namespace: kube-system
stringData:
  access-token: "3e8[...]ec5"
```

pvc.yaml:

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: jenkins-data
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  storageClassName: do-block-storage
```

* CSI Version:

1.3.0

* Kubernetes Version:

1.17

* Cloud provider/framework version, if applicable (such as Rancher):

OKD 4.5

max3903:
Other information:

I am using OKD 4.5 on Fedora CoreOS 31.

Here is the error when installing the CSI driver:

```
CustomResourceDefinition.apiextensions.k8s.io "volumesnapshots.snapshot.storage.k8s.io" is invalid: spec.version: Invalid value: "v1alpha1": must match the first version in spec.versions
```

The pod `csi-do-controller-0` remains in status `CrashLoopBackOff`.

4 out of 5 containers are in state `Running` but have this error message in the log:
```
connection.go:170] Still connecting to unix:///var/lib/csi/sockets/pluginproxy/csi.sock
```

The last one `csi-do-plugin` (digitalocean/do-csi-plugin:v1.3.0) remains in state `Waiting` and the logs says:
```
[...]

PR #333, opened 2020-07-04
Drop privileged flag

timoreimann:
This won't be needed anymore (as indicated in https://github.com/digitalocean/csi-digitalocean/issues/328#issuecomment-653712642).


timoreimann:
Turns out the flag is needed on the Node service in order to use mount propagation.

PR #339, opened 2020-09-05
Update csi-test to v4.0.1

timoreimann:
This change updates csi-test (aka csi-sanity) from our v3 fork to upstream v4.0.1. The test updates necessitate a number of smallish modifications to our driver, namely:

- Implement `ControllerGetVolume` which is required by the new [CSI health check feature](kubernetes/enhancements#1077) available in CSI spec 1.3.0 that csi-test is now based on. We return `Unimplemented` for now since the feature is still marked alpha, and if/when implemented should be done so in a dedicated PR. (It is worth mentioning that `ControllerGetVolume` won't be called anyway because we are not returning the right capabilities newly added for this feature.)
- Use a custom ID generator returning all integers node IDs (i.e., droplet IDs), which allows us to return a proper error code when the node ID cannot be converted from integer to string.
- Fake multiple droplets, which is needed to prevent running into max-volumes-per-node limit problems now that the number of tests has increased and a single (fake) node does not suffice anymore.
- Implement missing error handling in `NodeExpandVolume` which requires additional dependency injection so that we can mimic the right behavior for the tests.


adamwg (approved):
lgtm.

PR #340, opened 2020-09-07
Ensure target and staging paths are deleted on test start

timoreimann:
While csi-test / csi-sanity ensures that the paths are deleted after test completion, the cleanup cannot apply when tests fail hard (e.g., due to a panic). Thus, we do an extra cleanup step at test start time to ensure the test suite does not fail due to left-over files.


adamwg (approved):
lgtm.

PR #341, opened 2020-09-11
Update Kubernetes dependency to 1.19.1 and sidecars to latest

adamwg:
Update our Kubernetes dependencies to 1.19.1. While we're here, update all the sidecars to their latest versions:

* Provisioner 2.0.1 - requires an RBAC update for new permissions.
* Attacher 3.0.0 - requires an RBAC update for new permissions.
* Resizer 1.0.0 - requires the new commandline arg `--handle-volume-inuse-error=false` since we can handle online resizes.
* Registrar 2.0.1


timoreimann (commented):
This looks good. A few remarks / questions:

1. Strictly speaking, we do not need the flag to set ext4 since our driver implementation [is already assuming ext4 if no explicit driver is set](https://github.com/digitalocean/csi-digitalocean/blob/050726c39dd482325e7c75ca431f77bf4c0a714c/driver/node.go#L109-L112). Shouldn't make a difference, though less parameters to maintain may seem better.

1. The [external-provisioner change log](https://github.com/kubernetes-csi/external-provisioner/blob/28975c52f2f3f48f2b907cb828a57f7be0b3b0f2/CHANGELOG/CHANGELOG-2.0.md#changelog-since-v160) says the following:

>Add VolumeAttachment Lister to prevent calling DeleteVolume on the CSI plugin for volumes that are still attached to a kubernetes node. New RBAC rules for listing VolumeAttachment objects are required.

It sounds like we would need to extend the RBAC rules then?

2. Similar remark on [the csi-attacher change log:](https://github.com/kubernetes-csi/external-attacher/blob/bacf6647ca5139e17ea8deb759e260832ec716a6/CHANGELOG/CHANGELOG-3.0.md#changelog-since-v220):

>RBAC policy was updated to allow the external-attacher to patch VolumeAttachment.Status.

Could you also add a change log entry? 🙏

timoreimann:
It also seems like the images are being moved to k8s.gcr.io now. Pulling from Quay still works though, so maybe that's to support a transitioning period.

adamwg:
Ah, I had updated the RBAC rules elsewhere for testing, but forgot to apply those changes here. Added.

Also updated the changelog, and got rid of the `--default-fstype` as suggested.

timoreimann:
Looking good. 👍 

The CI is unhappy for what I believe to be reasons unrelated to your PR. (Other open, much more trivial PRs are failing too.) I'm investigating.

timoreimann:
CI is green. 🎉

timoreimann (approved):

timoreimann:
We may want to bump the Kubernetes deps to 1.19.2 now that it's out, just so we're on the latest.

PR #342, opened 2020-09-12
Fix built-in release version

timoreimann:
In #274, we started to release container images by promoting/retagging the latest `:master` image. However, this did not yield an updated version string baked into the binary; instead, we ended up with the _latest_ version used when we build the latest master.

This change modifies the release Github Action by recompiling and packaging the binary so that the right version is included.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ endif
 
 all: check-unused test
 
-publish: compile build push clean
+publish: clean compile build push
 
 .PHONY: update-k8s
 update-k8s:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -44,7 +44,7 @@ endif
 
 all: check-unused test
 
-publish: compile build push clean
+publish: clean compile build push
```
timoreimann, line 47: Drive-by improvement: cleaning after publishing didn't really make sense to me.

MorrisLaw (approved):

PR #347, opened 2020-09-24
Clean up clusters on PR completion

timoreimann:
This change moves the cleanup of clusters into a dedicated Github Actions workflow executed when a PR is completed (either through a merge or by being closed), allowing failed end-to-end test runs to be inspected.

The actual cleanup is carried out by a small script that we bundle with the runner image (as that one already comes with `doctl`). It depends on clusters having been tagged with a branch identifier, for which we leverage the cluster name suffix already employed to guarantee unique cluster names.

### DIFF
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -1,0 +1,24 @@
+name: cleanup-clusters
+
+env:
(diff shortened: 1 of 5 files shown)


timoreimann:
Closing PR temporarily to test if cleanup is working.

PR #351, opened 2020-10-04
Reclaim unused local volumes space

varshavaradarajan:
Delete unused local volumes if present. As we increase building images for newer versions, clearing up space as much as possible makes sense so as to stay within the limits for GH hosted runners


varshavaradarajan:
Opened PR from my fork. 😞

PR #352, opened 2020-10-04
Reclaim unused local volumes space

varshavaradarajan:
Delete unused local volumes if present. As we increase building images for newer versions, clearing up space as much as possible makes sense so as to stay within the limits for GH hosted runners

### DIFF
--- a/.github/workflows/delete.yaml
+++ b/.github/workflows/delete.yaml
@@ -16,5 +16,5 @@ jobs:
           BRANCH=$(echo -n ${BRANCH} | tr -c '[:alnum:]._-' '-')
           TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username": "'${DOCKER_USER}'", "password": "'${DOCKER_PASS}'"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
           images=("${BRANCH}-latest" "${BRANCH}-runtime" "${BRANCH}-tools" "${BRANCH}-tests-1.18" "${BRANCH}-tests-1.17" "${BRANCH}-builder")
-          for i in ${images[*]}; do curl -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/repositories/digitalocean/k8s-e2e-test-runner/tags/$i/; done
-          curl -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/repositories/digitalocean/do-csi-plugin-dev/tags/${BRANCH}/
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -69,6 +69,7 @@ jobs:
           sudo apt clean
           docker rmi $(docker image ls -aq)
           df -h
+          docker volume prune -f
```
timoreimann, line 72: nit: could you move it above the `df -h` line so we see the total disk space available after we also pruned the volumes?

timoreimann (approved):
Good catch -- didn't realize the workers came volumes as well.

LGTM. Left a small nit.

.github/workflows/test.yaml
varshavaradarajan: Changed the command to `docker system prune -af` and removed `docker rmi $(docker image ls -aq)` - system prune removes all unused data.

timoreimann: According to https://docs.docker.com/engine/reference/commandline/system_prune/, we need to pass `--volumes` to the prune command to include volumes.

varshavaradarajan: Done. Thanks 💙

.github/workflows/delete.yaml
```diff
@@ -16,5 +16,5 @@ jobs:
           BRANCH=$(echo -n ${BRANCH} | tr -c '[:alnum:]._-' '-')
           TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username": "'${DOCKER_USER}'", "password": "'${DOCKER_PASS}'"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
           images=("${BRANCH}-latest" "${BRANCH}-runtime" "${BRANCH}-tools" "${BRANCH}-tests-1.18" "${BRANCH}-tests-1.17" "${BRANCH}-builder")
-          for i in ${images[*]}; do curl -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/repositories/digitalocean/k8s-e2e-test-runner/tags/$i/; done
-          curl -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/repositories/digitalocean/do-csi-plugin-dev/tags/${BRANCH}/
+          for i in ${images[*]}; do curl --fail -s -X DELETE -H "Authorization: JWT ${TOKEN}" https://hub.docker.com/v2/repositories/digitalocean/k8s-e2e-test-runner/tags/$i/; done
```
timoreimann, line 19: If we use `-s` (`--silent`), then I think we should also add `-S` (or `--show-error`) so that error messages are not hidden. (See also [the curl docs on this parameter](https://curl.haxx.se/docs/manpage.html#-S).)

Similar below.

PR #358, opened 2020-10-15
Allow releasing from tag in Github Actions

timoreimann:
Our Makefile contains an (ugly) chain of conditions to prevent releasing anything but the `dev` tag from a non-master branch. This breaks our release after we moved to rebuilding instead of just retagging (#342).

We solve the matter by requiring a semver `$VERSION` in the `push` target, which is exactly what our release action already provides.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -47,7 +47,7 @@ jobs:
       - name: Build and push released plugin image
         run: |
           VERSION="$(git describe --tags)"
-          DOCKER_REPO=${DOCKER_ORG}/do-csi-plugin make publish
+          I_KNOW_WHAT_IAM_DOING=1 DOCKER_REPO=${DOCKER_ORG}/do-csi-plugin make publish
 
       - name: create Github release
         uses: softprops/action-gh-release@b7e450d  # v0.1.5
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,8 +23,7 @@ on:
   push:
     tags-ignore:
       # The tags below are for Kubernetes <1.14 which are not supported by the
-      # upstream end-to-end tests. We should find a different way to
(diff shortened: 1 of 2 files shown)


varshavaradarajan:
@timoreimann,
According to https://github.com/digitalocean/csi-digitalocean/blob/dd6772440daf1e06a6a0416e382c5dea59cbbd0f/.github/workflows/release.yaml#L15, the release workflow should get triggered on release of a tag only right? What is the value of  the `BRANCH` env var when this happens? It doesn't look like we are setting the BRANCH env var, so I would assume that it would be master?

And what exactly is checked out at the `checkout` step when this workflow triggers once a tag is pushed?

timoreimann:
>the release workflow should get triggered on a tag being pushed only, right?

That's correct: Github Actions checks out the commit backed by the tag that triggered the workflow.

>What is the value of BRANCH when this happens? It doesn't look like we are setting the BRANCH anywhere in the workflow, so I would assume that it would be master when git rev-parse --abbrev-ref HEAD is executed?

`git rev-parse --abbrev-ref HEAD` yields `HEAD` when executed from a checked out branch.

Your comment made me reconsider the guard we have in the `push` Makefile target: I think it actually does not make a lot of sense anymore to look at branches when we deploy from tags only. Even for the theoretical case where we release locally for whatever reason (there should really be none justifying that), it shouldn't happen from a commit that isn't a tag. (The only exception being that the repo isn't `digitalocean/do-csi-plugin`, e.g., we're deploying to personal repos for testing purposes.)
Therefore, I went ahead and changed the guard to check `$VERSION` for semver compliance and return an error if that fails. This should work from the release Action out-of-the-box since we're pre-populating that variable from `git describe --tags` (which yields the git tag, e.g., vX.Y.Z).

Let me know what you think.

varshavaradarajan (approved):

PR #359, opened 2020-10-15
Match semver without bash

timoreimann:
Makefile and Github Actions provides `/bin/sh` by default which does not support `[[`. We use grep instead which is just as good as more easily available.


varshavaradarajan (approved):
🚀

timoreimann:
Going renegade and merge right away to test the release.

PR #361, opened 2020-10-29
e2e: run tests against 1.19 minor version

timoreimann:
Now that 1.19 is publicly released, we can reference the minor version and let the DOKS service resolve it to the latest patch release.


waynr (approved):

gottwald (approved):
LGTM

PR #367, opened 2020-11-23
Update Go to 1.15.5

timoreimann:


.github/workflows/test.yaml
```diff
@@ -39,9 +39,9 @@ jobs:
           fetch-depth: 1
 
       - name: Go setup
-        uses: actions/setup-go@8a3a76c  # v2.1.2
+        uses: actions/setup-go@37335c7  # v2.1.3
```
gottwald, line 42: Why don't we just use the tag `v2.1.3` here?

timoreimann: It's a security best practice that prevents malicious attackers from running privileged code. See #343 too.

gottwald: Ah gotcha, because they're not under our control. SGTM

gottwald (approved):

.github/workflows/test.yaml
gottwald: Although, just to make sure: This would only protect accidental re-tagging IMHO.
A malicious one could easily use `gitbrute` to reproduce the same short-hash with a different content.
We'd need to use the long form of hashes to be safe here.

timoreimann: Good catch (and TIL about `gitbrute`). Will move to long commits in a subsequent PR.

timoreimann: #369 is that PR now.

PR #372, opened 2020-12-30
Add configuration options to support different driver modes

r0kas:
We have a use case where we want to run DO CSI driver not on a Droplet and just for Controller operations.
For creating/deleting volumes with DO API but not mounting them to specific droplet.

In this PR you'll find the solution which works for our use case at the moment.
Would like your feedback if it would possible to have such functionality in official project if yes, how far is this PR from how you would like to implement his?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 cmd/do-csi-plugin/do-csi-plugin
 .idea/
 *.sedbak
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -127,12 +127,19 @@ The suggested release manifests provide separate deployments for controller and 
 
 When running outside of DigitalOcean droplets, the driver can only function in **controller mode**.
 This requires to set the `--region` flag to a valid DigitalOcean region slug in addition to the other flags.
-The `--region` flag should **not** be set when running the driver on DigitalOcean droplets.
+
+The `--region` flag **must not** be set when running the driver on DigitalOcean droplets.
 
 Alternatively driver can be run in **node only mode** on DigitalOcean droplets.
 Driver would only handle node related requests like mount volume. Driver runs in **node only mode** when `--token` flag is not provided.
 
 Skip secret creation (section 1. in following deployment instructions) when using **node only mode** as API token is not required.
+
+| Modes                                     |  `--token` flag  |  `--region` flag |
+|-------------------------------------------|:----------------:|:----------------:|
+| Controller and Node mode in DigitalOcean  |:white_check_mark:|        :x:       |
+| Controller only mode not in DigitalOcean  |:white_check_mark:|:white_check_mark:|
+| Node only mode in DigitalOcean            |        :x:       |        :x:       |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -118,6 +118,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
 The [DigitalOcean Kubernetes](https://www.digitalocean.com/products/kubernetes/) product comes with the CSI driver pre-installed and no further steps are required.
 
 ---
+**Driver modes:**
+
+By default, driver runs in both controller and node mode. It would create disk Volumes on DigitalOcean infrastructure and mount them on the required node.
+
+Driver can be run in **controller only mode** outside DigitalOcean droplets.
+To use this mode `--region` flag (valid DigitalOcean region slug) must be provided together with `--token` flag (DigitalOcean API token).
+
+Alternatively driver can be run in **node only mode** on DigitalOcean droplets. Driver would only handle node related requests like mount volume.
+To us this mode `--region` and `--token` flags must not be provided.
+
+Skip secret creation (section 1. in following deployment instructions) when using **node only mode** as API token is not required.
+
+---
 
 **Requirements:**
 
--- a/README.md
+++ b/README.md
@@ -129,7 +129,8 @@ When running outside of DigitalOcean droplets, the driver can only function in *
 This requires to set the `--region` flag to a valid DigitalOcean region slug in addition to the other flags.
 The `--region` flag should **not** be set when running the driver on DigitalOcean droplets.
 
-Alternatively driver can be run in **node only mode** on DigitalOcean droplets. Driver would only handle node related requests like mount volume.
+Alternatively driver can be run in **node only mode** on DigitalOcean droplets.
+Driver would only handle node related requests like mount volume. Driver runs in **node only mode** when `--token` flag is not provided.
 
 Skip secret creation (section 1. in following deployment instructions) when using **node only mode** as API token is not required.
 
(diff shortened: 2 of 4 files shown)


timoreimann (commented):
👋 thanks for your PR. I'm fine with the idea to support running the controller outside of a DO droplet.

I left a few comments, mostly around suggestions to reduce the number of new flags we need. Let me know what you think.

timoreimann:
Would also be great to add a bit of documentation to the README explaining how the `region` flag is to be used when running the controller on a non-droplet host.

r0kas:
Sorry for coming back just now, had my github notification misconfigured... 
Updated the code with your suggestions!

cmd/do-csi-plugin/main.go
```diff
@@ -45,7 +46,11 @@ func main() {
 		os.Exit(0)
 	}
 
-	drv, err := driver.NewDriver(*endpoint, *token, *url, *doTag, *driverName, *debugAddr)
+	if *token == "" && *region != "" {
+		log.Fatalln("region flag must not be set when driver is running in node mode (with unset token flag)")
```
timoreimann, line 50:
```suggestion
		log.Fatalln("region flag must not be set when driver is running in node mode (i.e., token flag is unset)")
```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
 
 The [DigitalOcean Kubernetes](https://www.digitalocean.com/products/kubernetes/) product comes with the CSI driver pre-installed and no further steps are required.
 
+---
+**Driver modes:**
+
+By default, driver runs in both controller and node mode. It would create disk Volumes on DigitalOcean infrastructure and mount them on the required node.
```
timoreimann, line 123:
```suggestion
By default, the driver supports both the [controller and node mode](https://kubernetes-csi.github.io/docs/deploying.html): it can manage DigitalOcean Volumes via the cloud API and mount them on the required node. The actually used mode is determined by how the driver is deployed and configured.
```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+
+Driver can be run in **controller only mode** outside DigitalOcean droplets.
```
timoreimann, line 125:
```suggestion
When running outside of DigitalOcean droplets, the driver can only function in **controller mode**. This requires to set the `--region` flag (a valid DigitalOcean region slug) in addition to the other flags. The `--region` flag should **not** be set when running the driver on DigitalOcean droplets.
```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+To use this mode `--region` flag (valid DigitalOcean region slug) must be provided together with `--token` flag (DigitalOcean API token).
```
timoreimann, line 126:
```suggestion

```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+
+Alternatively driver can be run in **node only mode** on DigitalOcean droplets. Driver would only handle node related requests like mount volume.
```
timoreimann, line 128:
```suggestion
When running in **node mode only** on DigitalOcean droplets, the `token` flag is not needed (and consequently, no API token secret needs to be deployed). Additionally, in this case the `--region` and `--token` flags must not be specified together.
```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+To us this mode `--region` and `--token` flags must not be provided.
```
timoreimann, line 129:
```suggestion
```

README.md
```diff
@@ -117,6 +117,19 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+
+Skip secret creation (section 1. in following deployment instructions) when using **node only mode** as API token is not required.
```
timoreimann, line 131: Please drop the sentence as I merged it with the upper paragraph.

timoreimann (commented):
This is looking great! I just left a few more minor suggestions for improved wording / phrasing. The functionality looks correct, I'll run it through our e2e CI again. (Unfortunately, Github Actions does not have a great way to trigger the full suite from forks. 😞  )

timoreimann:
My manual e2e test runs were all successful 🎉 

This is ready to merge as soon as the last comments have been addressed.

r0kas:
Updated the phrasing. Thanks for spending the time on it!

.gitignore
```diff
@@ -1,3 +1,4 @@
 cmd/do-csi-plugin/do-csi-plugin
 .idea/
 *.sedbak
+.DS_Store
```
timoreimann, line 4: Could you drop this one again please? IMO it should be in your global .gitignore.

README.md
timoreimann: Does this suggestion still need to be applied?

README.md
timoreimann: Can we? :)

timoreimann (commented):
Some last left-overs (unless I missed something), then we're good I think. :-)

.gitignore
timoreimann: Actually, given that we also include `.idea`, I'm also okay with keeping it.

README.md
r0kas: Yeah, node mode does not interact with DigitalOcean API, it just performs node operations: mount, unmount, etc. On the fork I'm using it without token, thus secret is not required for this mode.

README.md
r0kas: I think yes, as it is possible to run driver in this mode if someone else would find a use case for it as we did.
Added additional sentence to be more clear how it's ran in this mode.

r0kas:
Added a table to readme modes paragraph which visualizes which fags should be set in which mode.

r0kas:
Ping

README.md
```diff
@@ -117,6 +117,30 @@ Kubernetes Release | DigitalOcean CSI Driver Version
...
+This requires to set the `--region` flag to a valid DigitalOcean region slug in addition to the other flags.
+
+The `--region` flag **must not** be set when running the driver on DigitalOcean droplets.
+
+Alternatively driver can be run in **node only mode** on DigitalOcean droplets.
+Driver would only handle node related requests like mount volume. Driver runs in **node only mode** when `--token` flag is not provided.
+
+Skip secret creation (section 1. in following deployment instructions) when using **node only mode** as API token is not required.
+
+| Modes                                     |  `--token` flag  |  `--region` flag |
```
timoreimann, line 138: This table is super useful, thanks a bunch 👍

timoreimann (approved):
This is looking great now. Thank you so much for your contribution 👍  -- merging now. 🎉

PR #380, opened 2021-01-20
Update sidecar versions and add 1.20 e2e tests

adamwg:
Update sidecar versions to the latest, staying on the 3.x shapshotter series since we don't support v1 snapshots yet and 4.x supports only Kubernetes 1.20+.

Add 1.20 to the e2e tests to validate that everything is working on 1.20 before cutting a release.


timoreimann:
Closing this PR because we (1) have (had) other sidecar-updating PRs by now with possibly more recent updates and (2) we cannot run on 1.20 as-is since that Kubernetes version's upstream e2e tests assume support of v1 snapshots, which we do not have yet.

PR #383, opened 2021-06-17
Update kubernetes dependencies to 1.21.1

adamwg:
As part of this change, use the resizefs implementation from `k8s.io/mount-utils`. This gets rid of the last reference to `k8s.io/kubernetes`, so we no longer need to vendor it :tada:.

While we're here, tidy up the go.mod file a bit by putting all the `replace`s in one stanza. I suspect we ended up with some separated out at the end via previous runs of update-k8s.sh, but I don't see any good reason to keep it that way.


go.mod
```diff
@@ -3,62 +3,60 @@ module github.com/digitalocean/csi-digitalocean
...
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.20.2
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.20.2
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.20.2
-	k8s.io/kubectl => k8s.io/kubectl v0.20.2
-	k8s.io/kubelet => k8s.io/kubelet v0.20.2
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.20.2
-	k8s.io/metrics => k8s.io/metrics v0.20.2
+	k8s.io/api => k8s.io/api v0.21.1
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.21.1
+	k8s.io/apimachinery => k8s.io/apimachinery v0.21.2-rc.0
```
gottwald, line 35: Is there a reason to use an `rc` release here? `v0.21.2` would also be available. Also applies to the other `rc`s in this file.

timoreimann: Probably the same reason as in https://github.com/digitalocean/csi-digitalocean/pull/318#discussion_r437012814.

gottwald: Seems like it. `v0.21.1` and `v0.21.2-rc.0` both point to `7559d78`

adamwg: Aha, I wondered about this as well but trusted that the tool was calculating things correctly. Updated to `v0.21.1` everywhere for consistency.

timoreimann (approved):

PR #389, opened 2021-08-23
bump dependencies to 0.21.3

varshavaradarajan:
```
GO111MODULE=on go get k8s.io/client-go@v0.21.3
GO111MODULE=on go get k8s.io/mount-utils@v0.21.3
GO111MODULE=on go mod tidy
GO111MODULE=on go mod vendor
```


varshavaradarajan:
@timoreimann , @MorrisLaw - I had opened a PR from my work with the same commit and the dockerhub secrets were not populated, so, delete and cleanup jobs failed. I tried rerunning, but it looks like they are still failing for the same. Is it alright if I merged this? I have to raise a PR for the changelog anyway, so that PR will do the cleanup.

PR #397, opened 2021-10-30
Support snapshots v1

timoreimann:
This also allows us to run against the recent upstream storage end-to-end tests again.

Remove the snapshot migration tool also as we stop supporting v1beta1 snapshots.


timoreimann:
This is now good for review and afterwards ready for merging.

gottwald (approved):

PR #407, opened 2021-12-17
Add snapshot validation webhook

timoreimann:
Add the snapshot validation webhook responsible for labeling invalid snapshots. This is the recommended approach in preparation to installing snapshots v1.

See [the upstream docs](https://github.com/kubernetes-csi/external-snapshotter/tree/138d310e5d2d102fcf90df96d7a8aaf6690ce76e#validating-webhook) for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,14 @@ default-token-jskxx   kubernetes.io/service-account-token   3         18h
 digitalocean          Opaque                                1         18h
 ```
 
-#### 2. Deploy the CSI plugin and sidecars
+#### 2. Provide authentication data for the snapshot validation webhook
+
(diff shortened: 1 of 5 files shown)


gottwald (approved):

varshavaradarajan:
@timoreimann - would this cause problems in legacy master droplet not being ready after an upgrade?

varshavaradarajan:
There have been some support issues where user configured webhooks with FailurePolicy `Fail` has caused problems right after an upgrade making the master droplet unready. Do we need some sort of check in the bootstrap-master script where initially, the FailurePolicy of this webhook will be `Ignore` and then after the master is ready, we can set it to `Fail`?

timoreimann:
@varshavaradarajan I think the kind of problems you are referring to are about failing web hooks that prevent critical components (e.g., Cilium) from coming up after an upgrade. In contrast, the snapshot validation webhook only applies to volume snapshot create and update requests, so it should not be a problem. At least, I cannot think of a scenario where a failing snapshot request could block provisioning, especially since the CSI driver should retry and eventually succeed once the web hook is up and running.

In the end though, it's up to the concrete user of our FOSS CSI driver to configure the web hook to their needs. As the changes to the README explain, the setup of the webhook in this PR is mostly to provide a (non-production-ready) example (and to enable easy end-to-end testing integration).

varshavaradarajan (approved):

PR #413, opened 2021-12-27
e2e: do not set platform to fix running on Mac

timoreimann:
Newer versions of Docker seem to fail when using the default platform since Darwin on amd64 has become a dedicated image architecture with Docker:

```
e2e_test.go:284: end-to-end tests failed: failed to create container: Error response from daemon: image with reference docker.io/digitalocean/k8s-e2e-test-runner:latest was found but does not match the specified platform: wanted darwin/amd64, actual: linux/amd64
```

Not setting the platform ensures the Linux image is requested.

### DIFF
--- a/test/e2e/docker_test.go
+++ b/test/e2e/docker_test.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containerd/containerd/platforms"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/mount"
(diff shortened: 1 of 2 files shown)


test/e2e/docker_test.go
```diff
@@ -51,7 +50,7 @@ type containerParams struct {
 // It returns an error when the container startup or cleanup fails, or when the
 // container returns a non-zero exit code.
 func runContainer(ctx context.Context, p containerParams) (retErr error) {
-	cli, err := client.NewEnvClient()
+	cli, err := client.NewClientWithOpts(client.FromEnv)
```
timoreimann, line 53: Drive-by improvement because the old method is deprecated.

varshavaradarajan (approved):

PR #420, opened 2022-01-10
Support v1.22 e2e

CollinShoop:
CON-4800 

- vendored 22.5 dependencies 
- add 1.22 version support
- upgrade base alpine from `3.11` -> `3.15` (optional, just keeping things up date)
- added missing controller `patch` permissions requirements that were missing [(see under breaking changes)](https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v5.0.0)
- added ginkgo skip for `[Feature:VolumeSourceXFS]` to avoid problematic tests - this should be removed after 1.22 whenever we have a fix for it. 

Vendor changes
```
go get k8s.io/apimachinery@v0.22.5
go get k8s.io/mount-utils@v0.22.5
go get k8s.io/utils@master
go mod tidy
go mod vendor
``

### CODE REVIEW

**Comment by timoreimann** [comment]
There are some conflicts here. @CollinShoop could it be that your branched off of master in your local checkout as opposed to the `support-v1-snapshots` branch? If the latter is picked as base, the number of conflicts should be minimal I think.

**Comment by CollinShoop** [comment]
> There are some conflicts here. @CollinShoop could it be that your branched off of master in your local checkout as opposed to the `support-v1-snapshots` branch? If the latter is picked as base, the number of conflicts should be minimal I think.

I did fork my branch off of `support-v1-snapshots` but I'm not sure what happened here. Just fixed it.

File: test/e2e/testdrivers/1.22.yaml [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+StorageClass:
+  FromName: true
+SnapshotClass:
+  FromName: true
+DriverInfo:
+  Name: dobs.csi.digitalocean.com-dev
+  SupportedSizeRange:
+    Max: 100Gi
+    Min: 1Gi
+  Capabilities:
```
CollinShoop, line 24: @timoreimann How do I know which Capabilities need to be provided?

timoreimann: That's a bit non-trivial, unfortunately: you basically need to know which capabilities are available in CSI (by reading up on the [CSI docs](https://kubernetes-csi.github.io/docs/) and the [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)), which of those our driver supports, and then take a good guess or look at the upstream e2e code to understand what the individual capabilities refer to exactly.

In this case, it's relatively easy since the caps don't seem to have changed between [1.21](https://github.com/kubernetes/kubernetes/blob/release-1.21/test/e2e/storage/framework/testdriver.go) and [1.22](https://github.com/kubernetes/kubernetes/blob/release-1.22/test/e2e/storage/framework/testdriver.go), so we should be fine copying what we have for 1.21 already.

CollinShoop: Ahh.. I was afraid of that. Guess I got lucky though 😅

test/e2e/run-versioned-e2e-tests.sh
```diff
@@ -68,5 +68,5 @@ if [[ "${SKIP_SEQUENTIAL_TESTS:-}" ]]; then
   echo 'Skipping sequential tests'
 else
   echo 'Running sequential tests'
-  ginkgo -v -focus="External.Storage${focus}.*(\[Feature:|\[Serial\])" -skip='\[Disruptive\]' "${E2E_TEST_FILE}" -- "-storage.testdriver=${TD_FILE}"
+  ginkgo -v -focus="External.Storage${focus}.*(\[Feature:|\[Serial\])" -skip='\[Disruptive\]|\[Feature:VolumeSourceXFS\]' "${E2E_TEST_FILE}" -- "-storage.testdriver=${TD_FILE}"
```
CollinShoop, line 71: Needed to avoid new tests that were added with 1.22 that are failing but also seem to fail on previous versions.

timoreimann (approved):

PR #440, opened 2022-06-22
CON-6258 Update CSI driver for Kubernetes 1.23

dylanrhysscott:
* vendored 23.7 dependencies
* add 1.23 version support
* upgrade base alpine from 3.15 -> 3.16 (optional, just keeping things up date)


test/e2e/testdrivers/1.23.yaml
```diff
@@ -0,0 +1,42 @@
+# Copyright 2020 DigitalOcean
```
timoreimann, line 1: nit: let's make this 2022 right away.

timoreimann (commented):
Looking really good. Need to make sure the CI passes now.

PR #441, opened 2022-06-24
Set storage size to minimum volume size

srikiz:
Setting required Bytes to minimum volume size (1Gi) in case it is less than the minimum volume size.

Tested locally and works as expected.

Input pvc & pod for test:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: csi-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1
  storageClassName: do-block-storage
---
kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app
spec:
  containers:
    - name: my-frontend
      image: busybox
      volumeMounts:
      - mountPath: "/data"
        name: my-do-volume
      command: [ "sleep", "1000000" ]
  volumes:
    - name: my-do-volume
      persistentVolumeClaim:
        claimName: csi-pvc

```
Logs for reference:

`
time="2022-06-24T16:23:44Z" level=warning msg="requiredBytes is less than minimum supported volume size, setting requiredBytes default to minimumVolumeSizeBytes" host_id=303920350 minimumVolumeSizeInBytes=1Gi region=blr1 requiredBytes=1 version=v4.1.0
`
```
kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM             STORAGECLASS       REASON   AGE
pvc-934a18e6-ab6e-47a9-8e4f-d736e75ca7e6   1Gi        RWO            Delete           Bound    default/csi-pvc   do-block-storage            4m17s
```

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -93,7 +93,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("volume capabilities cannot be satisified: %s", strings.Join(violations, "; ")))
 	}
 
-	size, err := extractStorage(req.CapacityRange)
+	size, err := d.extractStorage(req.CapacityRange)
 	if err != nil {
 		return nil, status.Errorf(codes.OutOfRange, "invalid capacity range: %v", err)
 	}
@@ -873,7 +873,7 @@ func (d *Driver) ControllerExpandVolume(ctx context.Context, req *csi.Controller
 		return nil, status.Errorf(codes.Internal, "ControllerExpandVolume could not retrieve existing volume: %v", err)
 	}
 
-	resizeBytes, err := extractStorage(req.GetCapacityRange())
+	resizeBytes, err := d.extractStorage(req.GetCapacityRange())
 	if err != nil {
 		return nil, status.Errorf(codes.OutOfRange, "ControllerExpandVolume invalid capacity range: %v", err)
 	}
@@ -936,7 +936,7 @@ func (d *Driver) ControllerGetVolume(ctx context.Context, req *csi.ControllerGet
 // range. If the capacity range is not satisfied it returns the default volume
 // size. If the capacity range is below or above supported sizes, it returns an
 // error.
-func extractStorage(capRange *csi.CapacityRange) (int64, error) {
+func (d *Driver) extractStorage(capRange *csi.CapacityRange) (int64, error) {
 	if capRange == nil {
 		return defaultVolumeSizeInBytes, nil
 	}
@@ -955,7 +955,11 @@ func extractStorage(capRange *csi.CapacityRange) (int64, error) {
 	}
 
 	if requiredSet && !limitSet && requiredBytes < minimumVolumeSizeInBytes {
-		return 0, fmt.Errorf("required (%v) can not be less than minimum supported volume size (%v)", formatBytes(requiredBytes), formatBytes(minimumVolumeSizeInBytes))
+		d.log.WithFields(logrus.Fields{
+			"requiredBytes":            formatBytes(requiredBytes),
+			"minimumVolumeSizeInBytes": formatBytes(minimumVolumeSizeInBytes),
+		}).Warn("requiredBytes is less than minimum supported volume size, setting requiredBytes default to minimumVolumeSizeBytes")
+		return minimumVolumeSizeInBytes, nil
 	}
 
 	if limitSet && limitBytes < minimumVolumeSizeInBytes {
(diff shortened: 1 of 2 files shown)


srikiz:
FYI - @timoreimann

driver/controller.go
```diff
@@ -955,7 +955,11 @@ func extractStorage(capRange *csi.CapacityRange) (int64, error) {
 	}
 
 	if requiredSet && !limitSet && requiredBytes < minimumVolumeSizeInBytes {
-		return 0, fmt.Errorf("required (%v) can not be less than minimum supported volume size (%v)", formatBytes(requiredBytes), formatBytes(minimumVolumeSizeInBytes))
+		d.log.WithFields(logrus.Fields{
+			"requiredBytes":            formatBytes(requiredBytes),
```
timoreimann, line 959: Would you mind using camel case here and on the field below in order to be consistent with the existing style?

driver/controller.go
```diff
@@ -955,7 +955,11 @@ func extractStorage(capRange *csi.CapacityRange) (int64, error) {
...
+			"minimumVolumeSizeInBytes": formatBytes(minimumVolumeSizeInBytes),
+		}).Warn("requiredBytes is less than minimum supported volume size, setting requiredBytes default to minimumVolumeSizeBytes")
+		return minimumVolumeSizeInBytes, nil
```
timoreimann, line 962: Nit: Let's say "minimum volume size" to not give the impression that `minimumVolumeSizeBytes` is a known value from the CSI spec (which `requiredBytes` is only).

timoreimann (commented):
Two small suggestions, otherwise all good. 👍

driver/controller_test.go
```diff
@@ -181,14 +181,20 @@ func (*fakeTagsDriver) UntagResources(context.Context, string, *godo.UntagResour
 }
 
 func TestControllerExpandVolume(t *testing.T) {
+	default_volume := &godo.Volume{
```
timoreimann, line 184: Please use snake case here, i.e., `defaultVolume`.

driver/controller_test.go
```diff
@@ -200,6 +206,10 @@ func TestControllerExpandVolume(t *testing.T) {
 		},
 		{
 			name: "requested size less than minimum supported size",
+			godo_volume: &godo.Volume{
+				ID:            "volume-id",
+				SizeGigaBytes: (minimumVolumeSizeInBytes / giB),
```
timoreimann, line 211: Should we subtract a bit to make sure we're really below the minimum size (and not hitting it exactly)?

driver/controller_test.go
```diff
@@ -181,14 +181,20 @@ func (*fakeTagsDriver) UntagResources(context.Context, string, *godo.UntagResour
...
+		name        string
+		req         *csi.ControllerExpandVolumeRequest
+		resp        *csi.ControllerExpandVolumeResponse
+		err         error
+		godo_volume *godo.Volume
 	}{
 		{
-			name: "request exceeds maximum supported size",
+			name:        "request exceeds maximum supported size",
+			godo_volume: default_volume,
```
timoreimann, line 197: Instead of requiring to define the volume for every current and future test case, could we make it optional and let the test body pick `defaultVolume` if the volume from the test case struct is nil?

driver/controller_test.go
```diff
@@ -181,14 +181,20 @@ func (*fakeTagsDriver) UntagResources(context.Context, string, *godo.UntagResour
...
 	tcs := []struct {
-		name string
-		req  *csi.ControllerExpandVolumeRequest
-		resp *csi.ControllerExpandVolumeResponse
-		err  error
+		name        string
+		req         *csi.ControllerExpandVolumeRequest
+		resp        *csi.ControllerExpandVolumeResponse
+		err         error
+		godo_volume *godo.Volume
```
timoreimann, line 193: I'd name this just `volume` since the type is embedded in the `*godo.Volume` type. This would also keep the diff in this PR slightly smaller.

driver/controller_test.go
srikiz: Sure, updated.

driver/controller_test.go
srikiz: Sure, done.

timoreimann:
Merging since PRs from forks still cannot leverage Github Actions secrets, and the unit tests should suffice here.

timoreimann (approved):

MattIPv4:
Hey, @srikiz - Thanks a ton for this sweet PR! 😄

Would you please shoot me an email when you get a chance?
mcowley at digitalocean dot com 🎉

PR #457, opened 2022-08-18
[CON-7035] Update CSI driver for Kubernetes 1.24

mo-rieger:
* add e2e test support for Kubernetes 1.24
* update Kubernetes dependencies to support 1.24


Makefile
```diff
@@ -145,6 +145,14 @@ runner-build:
 		-t $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)builder -f test/e2e/Dockerfile test/e2e
 
 
+	@echo "building target tests-1.24"
+	@docker build --target tests-1.24 \
+		--cache-from $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)builder \
+		--cache-from $(CANONICAL_RUNNER_IMAGE):builder \
+		--cache-from $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)tests-1.24 \
```
timoreimann, line 152: You need to include 1.24 in the targets below as well for caching purposes, similar to how, say, 1.23 is used multiple times.

test/e2e/testdrivers/1.24.yaml
```diff
@@ -36,7 +36,6 @@ DriverInfo:
     volumeLimits: true
     singleNodeVolume: true
     topology: true
-    capacity: true
```
timoreimann, line 39: Just to confirm: did you check if any new options were added to the upstream 1.24 configuration struct?

test/e2e/Dockerfile
```diff
@@ -56,8 +65,6 @@ ARG TINI_VERSION=0.18.0
 ARG DOCTL_VERSION=1.76.2
 # ginkgo is needed to run the upstream end-to-end tests.
 ARG GINKGO_VERSION=1.10.3
-# kubectl is needed by some tests.
-ARG KUBECTL_VERSION=1.23.7
```
timoreimann, line 60: Is this not needed anymore?

mo-rieger: I removed it, because it was not used. In l. 78ff. we just install the latest stable version.

timoreimann: 👍

test/e2e/testdrivers/1.24.yaml
mo-rieger: Thanks for the hint. There are two options that we are not using atm.
```
	CapVolumeMountGroup   Capability = "volumeMountGroup"   // Driver has the VolumeMountGroup CSI node capability. Because this is a FSGroup feature, the fsGroup capability must also be set to true.

	// The driver publishes storage capacity information: when the storage class
	// for dynamic provisioning exists, the driver is expected to provide
	// capacity information for it.
	CapCapacity Capability = "capacity"
```
Do we want to enable these capabilities?

timoreimann: Looking at the corresponding CSI documentation pages for [CSI Driver fsGroup Support](https://kubernetes-csi.github.io/docs/support-fsgroup.html) and [Storage Capacity Tracking](https://kubernetes-csi.github.io/docs/storage-capacity-tracking.html), it looks to me like we'd need to do some work to support these features. I think it's fine to keep them disabled for now.

Would you mind setting them to `false` explicitly to serve as a future note/reminder for us what the level of support for our driver currently is?

khayyamsaleem (approved):

PR #461, opened 2022-10-14
Update k8s deps to 1.25

varshavaradarajan:
Really nice instructions in the tests/e2e/README.md! Thank you!


Makefile
```diff
@@ -144,6 +146,13 @@ runner-build:
...
 
+	@echo "building target tests-1.25"
+	@docker build --target tests-1.25 \
+		--cache-from $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)builder \
+		--cache-from $(CANONICAL_RUNNER_IMAGE):builder \
+		--cache-from $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)tests-1.25 \
+		--cache-from $(CANONICAL_RUNNER_IMAGE):tests-1.25 \
+		-t $(RUNNER_IMAGE):$(RUNNER_IMAGE_TAG_PREFIX)tests-1.25 -f test/e2e/Dockerfile test/e2e
 
 	@echo "building target tests-1.24"
```
timoreimann, line 157: This and all following `tests-*` builds should now reference the 1.25 cache. (See also how that was the case for 1.23 and below when 1.24 was the latest build.)

test/e2e/testdrivers/1.25.yaml
```diff
@@ -0,0 +1,44 @@
...
+StorageClass:
+  FromName: true
+SnapshotClass:
+  FromName: true
+DriverInfo:
+  Name: dobs.csi.digitalocean.com-dev
+  SupportedSizeRange:
+    Max: 100Gi
+    Min: 1Gi
+  Capabilities:
```
timoreimann, line 24: Just to check: have you confirmed that no new capabilities were added which we may want to enable? (It's usually not the case since our driver would normally need an update, but worth checking anyway.)

test/e2e/Dockerfile
```diff
@@ -20,6 +20,15 @@ RUN apt-get update && \
...
 WORKDIR /go/src/k8s.io
 
+### Kubernetes 1.25
+FROM builder AS tests-1.25
+ARG KUBE_VERSION_1_25=1.25.2
+ARG KUBE_VERSION_1_25_E2E_BIN_SHA256_CHECKSUM=b9ec7494abed4d5d40f4027e4211da0a2b33f76f81f6813430a9ebd142ca83a0
+
+RUN curl --fail --location https://dl.k8s.io/v${KUBE_VERSION_1_25}/kubernetes-test-linux-amd64.tar.gz | tar xvzf - --strip-components 3 kubernetes/test/bin/e2e.test
+RUN echo "${KUBE_VERSION_1_25_E2E_BIN_SHA256_CHECKSUM}" e2e.test | sha256sum --check
+RUN cp e2e.test /e2e.1.24.test
```
timoreimann, line 30: This must say 1.25.

timoreimann (changes requested):

mo-rieger:
Hmm, for some reason, ginkgo tests immediately time out after the cluster is provisioned....digging into it

varshavaradarajan:
@mo-rieger : https://github.com/onsi/ginkgo/issues/1054

test/e2e/Dockerfile
```diff
@@ -20,52 +20,59 @@ RUN apt-get update && \
...
 
+### Kubernetes 1.25
+FROM builder AS tests-1.25
+ARG KUBE_VERSION_1_25=1.25.2
+ARG KUBE_VERSION_1_25_E2E_BIN_SHA256_CHECKSUM=b9ec7494abed4d5d40f4027e4211da0a2b33f76f81f6813430a9ebd142ca83a0
+
+RUN curl --fail --location https://dl.k8s.io/v${KUBE_VERSION_1_25}/kubernetes-test-linux-amd64.tar.gz | tar xvzf - --strip-components 3 kubernetes/test/bin/e2e.test kubernetes/test/bin/ginkgo
+RUN echo "${KUBE_VERSION_1_25_E2E_BIN_SHA256_CHECKSUM}" e2e.test | sha256sum --check
+RUN cp e2e.test /e2e.1.25.test
+RUN cp ginkgo /ginkgo-1.25
```
timoreimann, line 31: Given we now require Kubernetes version-specific ginkgo binaries, is there a need to extend our "how to support a new Kubernetes minor version" documentation in the repo?

timoreimann (commented):
One quick question, otherwise good.

test/e2e/Dockerfile
varshavaradarajan: Updated the e2e/Readme.md to say copy version specific ginkgo into final runner image.

timoreimann (approved):
LGTM assuming we can get the tests to pass.

varshavaradarajan:
I squashed commits and amended the last commit on the last push too quickly. Will rerun after a while.

PR #469, opened 2023-01-10
Update csi driver for k8s 1.26

llDrLove:
Followed the [Prepare CSI driver for a new kubernetes minor version](https://github.com/digitalocean/csi-digitalocean#prepare-csi-driver-for-a-new-kubernetes-minor-version) section for the 1.26 k8s release.

CLOSES: CON-8686


.github/workflows/delete.yaml
```diff
@@ -15,6 +15,6 @@ jobs:
         run: |
           BRANCH=$(echo -n ${BRANCH} | tr -c '[:alnum:]._-' '-')
           TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username": "'${DOCKER_USER}'", "password": "'${DOCKER_PASS}'"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
-          images=("${BRANCH}-latest" "${BRANCH}-runtime" "${BRANCH}-tools" "${BRANCH}-tests-1.25" "${BRANCH}-tests-1.24" "${BRANCH}-tests-1.23" "${BRANCH}-tests-1.22" "${BRANCH}-builder")
+          images=("${BRANCH}-latest" "${BRANCH}-runtime" "${BRANCH}-tools" "${BRANCH}-tests-1.26" "${BRANCH}-tests-1.25" "${BRANCH}-tests-1.24" "${BRANCH}-tests-1.23" "${BRANCH}-tests-1.22" "${BRANCH}-builder")
```
timoreimann, line 18: We can probably use the opportunity and get rid of 1.22; and while we're here remove 1.23 as well since it's already unsupported upstream, and DOKS will follow soonish.

(If this is not part of the documentation, could I ask you to add it, i.e., suggest to use the opportunity when we add support for a new version to also consider removing unsupported versions?)

test/e2e/testdrivers/1.26.yaml
```diff
@@ -0,0 +1,44 @@
...
+StorageClass:
+  FromName: true
+SnapshotClass:
+  FromName: true
+DriverInfo:
+  Name: dobs.csi.digitalocean.com-dev
+  SupportedSizeRange:
+    Max: 100Gi
+    Min: 1Gi
+  Capabilities:
```
timoreimann, line 24: Just to confirm: have we validated that no new capabilities were added in the e2e test suite/framework that we may possibly support already?

timoreimann (commented):
This is looking good. Left one suggestion to clean up and one question.

.github/workflows/delete.yaml
timoreimann: Decided out-of-band that this can be handled by a follow-up PR.

test/e2e/testdrivers/1.26.yaml
timoreimann: Discovered together that upstream testing didn't support any new capabilities that our driver would support already. (This is becoming increasingly unlikely now that the upstream tests support all baseline features.)

timoreimann (approved):

PR #474, opened 2023-01-11
Update CSI dependencies for k8s v1.26

llDrLove:


timoreimann (approved):
Approving already despite PR being in draft mode still since all updates are minor version or less only.

Please ping if another review is required.

PR #478, opened 2023-03-01
CON-9040 driver can accidentally format existing volume

llDrLove:
There is a race condition where DOBS volumes look to our indication method as if they'd fully be attached, but aren't completely.
The result can be a misinterpretation by CSI which causes a force-format of the volume.

This PR uses a new method to check the attachment of the volume by looking at the `running` state in the `/sys/class/block/<device name like sda>/device/state` file. This can be enabled using the new flag `--validate-attachment=true` flag.


driver/mounter.go
```diff
@@ -208,6 +214,50 @@ func (m *mounter) Unmount(target string) error {
 	return mount.CleanupMountPoint(target, m.kMounter, true)
 }
 
+func (m *mounter) IsRunning(source string, cmdContext execContext) bool {
+	cmd := fmt.Sprintf("ls -l %s | awk '{print $NF}'", source)
```
timoreimann, line 218: There's a lot of command execution involved which isn't ideal. I think we can get around it.

Starting here, I believe we can use https://pkg.go.dev/path/filepath#EvalSymlinks.

driver/mounter.go
```diff
@@ -208,6 +214,50 @@ func (m *mounter) Unmount(target string) error {
...
+
+	if err != nil {
+		m.log.WithFields(logrus.Fields{
+			"cmd": cmd,
+		}).Errorf("error retrieving the mount name, err: %v", err)
+
+		return false
+	}
+
+	mountName := strings.TrimSuffix(string(bytes.Replace(out, []byte("../../"), []byte(""), 1)), "\n")
```
timoreimann, line 229: I think using https://pkg.go.dev/path/filepath#Split would be cleaner here.

driver/mounter.go
```diff
@@ -208,6 +214,50 @@ func (m *mounter) Unmount(target string) error {
...
+
+	cmd = fmt.Sprintf("cat /sys/class/block/%s/device/state | grep -x running | wc -l", mountName)
```
timoreimann, line 231: I'd prefer if we read the file into memory and parsed it for the `running` field.

timoreimann (commented):
I forgot to mention the following: it'd be great if we could introduce a CLI flag that controls whether the new check is being used. That way, we can turn it off easily should anything go wrong or not work out the way we want it to be.

driver/mounter.go
```diff
@@ -214,47 +230,41 @@ func (m *mounter) Unmount(target string) error {
...
-	out, err := cmdContext("/bin/sh", "-c", cmd, source).CombinedOutput()
-
+func (m *mounter) IsRunning(av iAttachmentValidator, source string) bool {
+	out, err := av.evalSymlinks(source)
 	if err != nil {
 		m.log.WithFields(logrus.Fields{
-			"cmd": cmd,
-		}).Errorf("error retrieving the mount name, err: %v", err)
+			"source": source,
+		}).Errorf("error evaluating the sym link, err: %v", err)
```
timoreimann, line 238: The pattern of logging errors and returning `false` only to turn that outcome into another error at the call site seems cumbersome to me. I'd prefer if we just always returned an error with all relevant info contained. For instance, in this case we could return

```go
return fmt.Errorf("failed evaluating symlink %q: %s", source, err)
```

(Side-note: I used `%s` here because we want the string representation of an error with certainty.)

That way, the call site can do the logging centrally.

If we want to follow this path, then I think it'd also make sense to rename the method to something like `EnsureAttached` and let it return an `error` only.

driver/mounter.go
```diff
@@ -214,47 +230,41 @@ func (m *mounter) Unmount(target string) error {
...
 		return false
 	}
 
-	if strings.TrimSuffix(string(out), "\n") != "1" {
+	m.log.WithFields(logrus.Fields{
+		"source":      source,
+		"fileContent": string(fileContent),
+	}).Info("state file content")
+
+	if string(fileContent) != fmt.Sprintf("%s\n", runningState) {
```
timoreimann, line 259: Let's check DO-internally if this is the best way to check for the running state.

driver/mounter.go
```diff
@@ -214,47 +230,41 @@ func (m *mounter) Unmount(target string) error {
 	return mount.CleanupMountPoint(target, m.kMounter, true)
 }
 
-func (m *mounter) IsRunning(source string, cmdContext execContext) bool {
-	cmd := fmt.Sprintf("ls -l %s | awk '{print $NF}'", source)
-	out, err := cmdContext("/bin/sh", "-c", cmd, source).CombinedOutput()
-
+func (m *mounter) IsRunning(av iAttachmentValidator, source string) bool {
```
timoreimann, line 233: Can we make `iAttachmentValidator` a member/field of `mounter`? This would hide the dependency injection more elegantly and keep the `IsRunning` signature simpler.

driver/mounter_mock.go
```diff
@@ -0,0 +1,224 @@
...
+	recorder *MockMounterMockRecorder
+}
+
+// MockMounterMockRecorder is the mock recorder for MockMounter.
+type MockMounterMockRecorder struct {
+	mock *MockMounter
+}
+
+// NewMockMounter creates a new mock instance.
+func NewMockMounter(ctrl *gomock.Controller) *MockMounter {
```
timoreimann, line 79: Why do we need to mock the entire mounter? Shouldn't it suffice the mock the attachment validator interface?

timoreimann, line 79: As a side note, I'm personally more a fan of stubbing than mocking when possible since it's usually simpler (at the price of a bit more verbosity). For instance, our test attachment validator could look like this:

```go
type testAttachmentValidator struct {
  readFileFunc func(name string) ([]byte, error)
  evalSymlinksFunc func(path string) (string, error)
}

func (av *testAttachmentValidator) readFile(name string) ([]byte, error) {
  return av.readFileFunc(name)
}

func (av *testAttachmentValidator) evalSymlinks(path string) (string, error) {
  return av.evalSymlinksFunc(path)
}
```

This would seem sufficient as we do not need to verify that certain methods are being called -- it should conclude naturally from the fact that we have to rely on the interface and it being implemented appropriately.

Using a stub also means we don't need extra tooling and processes to update the mocks if the interface changes (which is something I'd otherwise ask to be documented in this PR).

driver/mounter_test.go
```diff
@@ -1,170 +1,80 @@
...
-
-	return cmd
-}
-
 func Test_mounter_IsRunning(t *testing.T) {
-	type fields struct {
-		log      *logrus.Entry
-		kMounter *mount.SafeFormatAndMount
-	}
 	type args struct {
```
timoreimann, line 11: Can we drop the `args` struct indirection to be more consistent with the pre-existing test style?

driver/mounter_test.go
```diff
@@ -1,170 +1,80 @@
...
-		source     string
-		cmdContext execContext
+		source string
 	}
 	tests := []struct {
 		name   string
-		fields fields
 		args   args
+		expect func(*MockiAttachmentValidator)
 		want   bool
```
timoreimann, line 18: Relying on a boolean expectation only opens the door for false positives: we could be returning false in a case where we did not expect it so (e.g., when the system-under-test is actually taking an unexpected code path that happens to yield the same result).

It's usually better to test against concrete errors (or substrings of error messages, which is okay to do in tests).

cmd/do-csi-plugin/main.go
```diff
@@ -40,6 +40,7 @@ func main() {
 		defaultVolumesPageSize = flag.Uint("default-volumes-page-size", 0, "The default page size used when paging through volumes results (default: do not specify and let the DO API choose)")
 		doAPIRateLimitQPS      = flag.Float64("do-api-rate-limit", 0, "Impose QPS rate limit on DigitalOcean API usage (default: do not rate limit)")
 		version                = flag.Bool("version", false, "Print the version and exit.")
+		validateAttachment     = flag.Bool("validate-attachment", false, "Validate if the attachment is in a running state.")
```
timoreimann, line 43: Similar to my other comment about hiding the "runningness" nature of a device, I'd suggest we formulate the description here as

```
Validate if the attachment has fully completed before formatting/mounting the device
```

(nit: no trailing punctuation.)

cmd/do-csi-plugin/main.go
```diff
@@ -40,6 +40,7 @@ func main() {
 		defaultVolumesPageSize = flag.Uint("default-volumes-page-size", 0, "The default page size used when paging through volumes results (default: do not specify and let the DO API choose)")
 		doAPIRateLimitQPS      = flag.Float64("do-api-rate-limit", 0, "Impose QPS rate limit on DigitalOcean API usage (default: do not rate limit)")
 		version                = flag.Bool("version", false, "Print the version and exit.")
```
timoreimann, line 42: Let's keep the `version` flag last in the list since it's usually the least important.

timoreimann:
Let's also make sure to document the new flag in the README.

driver/mounter.go
```diff
@@ -71,7 +87,7 @@ type Mounter interface {
 	Unmount(target string) error
 
 	// IsRunning checks whether the source device is in the running state.
-	IsRunning(source string, cmdContext execContext) bool
+	IsRunning(av iAttachmentValidator, source string) bool
```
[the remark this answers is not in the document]
llDrLove: I agree. I'll switch it to `IsAttached`.

driver/mounter.go
```diff
@@ -214,47 +230,41 @@ func (m *mounter) Unmount(target string) error {
...
+		}).Errorf("error evaluating the sym link, err: %v", err)
 
 		return false
 	}
 
-	mountName := strings.TrimSuffix(string(bytes.Replace(out, []byte("../../"), []byte(""), 1)), "\n")
-
-	cmd = fmt.Sprintf("cat /sys/class/block/%s/device/state | grep -x running | wc -l", mountName)
-	out, err = cmdContext("/bin/sh", "-c", cmd).Output()
+	_, file := filepath.Split(out)
```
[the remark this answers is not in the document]
llDrLove: Sure. I've refactored it to `deviceName`.

cmd/do-csi-plugin/main.go
llDrLove: Removed the trailing punctuation and used the above message.

driver/mounter_test.go
llDrLove: I've refactored the unit tests to be in the same format as the `health_test.go` tests.

driver/mounter_test.go
llDrLove: I've changed the method to return a bool and and error instead to tackle what you are describing above.

driver/mounter_mock.go
llDrLove: Sure. I've refactored it to use your struct. It's cleaner and one dependency/package less to maintain.

driver/mounter.go
llDrLove: Nice ! I've refactored all the errors to use the new proposed format.

driver/mounter.go
```diff
@@ -208,6 +232,26 @@ func (m *mounter) Unmount(target string) error {
...
+	}
+
+	_, deviceName := filepath.Split(out)
+	deviceStateFilePath := fmt.Sprintf("/sys/class/block/%s/device/state", deviceName)
+	deviceStateFileContent, err := m.attachmentValidator.readFile(deviceStateFilePath)
+	if err != nil {
+		return false, fmt.Errorf("error reading the device state file %q: %s", deviceStateFilePath, err)
+	}
+
+	if string(deviceStateFileContent) != fmt.Sprintf("%s\n", runningState) {
```
timoreimann, line 248: This could probably be a tad more robust if we used [`strings.TrimSpace()`](https://pkg.go.dev/strings#TrimSpace), allowing us to not care what the newline character implementation is.

timoreimann, line 248: It'd be nice if we returned the content of the file when it does not match to ease troubleshooting. Returning an error for this case as well would achieve that goal (which would then also render the boolean return type obsolete).

Alternatively, we can pass in a logger for this scenario.

I am slightly preferring the error approach since it seems a bit easier to use and test against, but no strong opinion. Will leave the final decision to you.

driver/mounter_test.go
```diff
@@ -0,0 +1,110 @@
...
+			attachmentValidator: &testAttachmentValidator{
+				readFileFunc: nil,
+				evalSymlinksFunc: func(path string) (string, error) {
+					return "", errors.New("error")
+				},
+			},
+		}
+
+		_, err := m.IsAttached("")
+		if err == nil {
```
timoreimann, line 38: We should check for a specific error string (`"error"` in this case) so avoid false positives as I mentioned in a previous comment.

Also applies for other test cases.

driver/mounter_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (av *testAttachmentValidator) evalSymlinks(path string) (string, error) {
+	return av.evalSymlinksFunc(path)
+}
+
+func Test_mounter_IsAttached(t *testing.T) {
+	t.Run("could not evaluate the symbolic link", func(t *testing.T) {
+		m := &mounter{
+			log:      logrus.NewEntry(logrus.New()),
+			kMounter: nil,
```
timoreimann, line 28: `nil` is the default value for the type of the `kMounter` variable, which means we don't need to specify it in any test case.

driver/mounter_test.go
```diff
@@ -0,0 +1,110 @@
...
+			attachmentValidator: &testAttachmentValidator{
+				readFileFunc: nil,
```
timoreimann, line 30: If we stub this function out to return a different error, then it becomes clear how the test failed our implementations happens to accidentally call the wrong one. Right now, if we were to hit `readFileFunc`, the code would panic, which is harder to debug.

(I know that the code currently is simple enough for this to be very unlikely, but it's a good practice to account for the (test) maintainability of possible extensions, e.g., refactorings/additions to the logic.)

driver/mounter_test.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+		}
+
+		isAttached, err := m.IsAttached("")
+		if err != nil {
+			t.Errorf("expected no error: %s", err)
+		}
+
+		if isAttached {
+			t.Errorf("IsAttached() must return false when the state file does not contains the content: %s\n", runningState)
```
timoreimann, line 83: Unlike `fmt.*[pP]rintf`, a newline is not needed at the end of `t.Errorf` and `t.Fatalf`.

Also applies elsewhere in this file.

driver/mounter_test.go
```diff
@@ -0,0 +1,110 @@
...
+func (av *testAttachmentValidator) readFile(name string) ([]byte, error) {
+	return av.readFileFunc(name)
+}
+
+func (av *testAttachmentValidator) evalSymlinks(path string) (string, error) {
+	return av.evalSymlinksFunc(path)
+}
+
+func Test_mounter_IsAttached(t *testing.T) {
+	t.Run("could not evaluate the symbolic link", func(t *testing.T) {
```
timoreimann, line 25: When I asked to inline `args` in a previous comment, I should have probably been clearer in saying that the anonymous `struct` slice is okay to keep in order to avoid code repetitions (which we currently have a bit in this case).

Would suggest something like this:

```go
tests := []test{
  name string
  source string
  av AttachmentValidator
  errorMsg string
  wantResult bool
}{
  {
    [...]
  }
}

for _, tt := range tests {
  t.Run(tt.name, func(t *testing.T) {
    // Set up fake mounter.
    gotResult, err := m.IsAttached(tt.source)
    if tt.errorMesg != "" {
      assert.ErrorContains(t, err, tt.errorMsg)
    } else {
      assert.Equal(t, gotResult, tt.wantResult)
    }
  })
}
```

This style of testing is a pattern we use very often.

The assertions come from the [gotest.tools library](https://pkg.go.dev/gotest.tools/v3) which I recently discovered and personally like very much. I am on a mission to introduce it wherever I can. :)

driver/node.go
```diff
@@ -130,6 +130,13 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	if noFormat {
 		log.Info("skipping formatting the source device")
 	} else {
+		if d.validateAttachment {
+			isAttached, err := d.mounter.IsAttached(source)
+			if !isAttached || err != nil {
```
timoreimann, line 135: Using a logical operator could be quite confusing to the user because the `err` variable in the output below can be nil.

Let's separate this out into a check for `err != nil` and another one for `!isAttached` with distinct (error/log) messages.

driver/mounter.go
```diff
@@ -102,8 +125,9 @@ func newMounter(log *logrus.Entry) *mounter {
 	}
 
 	return &mounter{
-		kMounter: kMounter,
-		log:      log,
+		kMounter:            kMounter,
+		log:                 log,
+		attachmentValidator: &(prodAttachmentValidator{}),
```
timoreimann, line 130: The parentheses `()` should not be required.

driver/mounter.go
```diff
@@ -208,6 +232,26 @@ func (m *mounter) Unmount(target string) error {
 	return mount.CleanupMountPoint(target, m.kMounter, true)
 }
 
+func (m *mounter) IsAttached(source string) (bool, error) {
+	out, err := m.attachmentValidator.evalSymlinks(source)
+	if err != nil {
+		return false, fmt.Errorf("error evaluating the symbolic link %q: %s", source, err)
+	}
+
+	_, deviceName := filepath.Split(out)
```
timoreimann, line 241: Just to be safe, can we have an explicit check if `deviceName` is non-empty and error out with a clear error message if it is?

timoreimann:
Forgot to mention: we can already extend the CHANGELOG file for this feature.

driver/mounter.go
llDrLove: Your proposition is cleaner indeed. Adding it.

llDrLove: I'll return an error with the following message to ease troubleshooting. 

```go 
return false, fmt.Errorf("error comparing the state file content, expected: %s, got: %s", runningState, string(deviceStateFileContent))
```

driver/mounter_test.go
llDrLove: True. I removed it everywhere in this file.

driver/mounter_test.go
llDrLove: Refactored the tests with the above.

driver/mounter_test.go
llDrLove: True but I refactored all the tests to a table driven test suite and this comment does not apply anymore.

driver/node.go
llDrLove: Separated the conditions and added a distinct log message to improve  debuggability

driver/mounter.go
llDrLove: Removed.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## unreleased
 
+## v4.5.1 - 2023.03.03
```
timoreimann, line 3: Please omit the header -- that's going to be added at the time we do the release for sure.

driver/mounter.go
```diff
@@ -208,6 +232,30 @@ func (m *mounter) Unmount(target string) error {
...
+	deviceStateFileContent, err := m.attachmentValidator.readFile(deviceStateFilePath)
+	if err != nil {
+		return false, fmt.Errorf("error reading the device state file %q: %s", deviceStateFilePath, err)
+	}
+
+	if string(deviceStateFileContent) != strings.TrimSpace(runningState) {
+		return false, fmt.Errorf("error comparing the state file content, expected: %s, got: %s", runningState, string(deviceStateFileContent))
+	}
+
+	return true, nil
```
timoreimann, line 256: At this point, we only return `true` when `err == nil`, which means we could drop the boolean return value and rely on the error return value exclusively.

driver/mounter_test.go
```diff
@@ -0,0 +1,137 @@
...
+func Test_mounter_IsAttached(t *testing.T) {
+	testSource := "test-source"
+	testEvalSymlinkErr := errors.New("eval sym link err")
+	testReadFileErr := errors.New("read file err")
+
+	tests := []struct {
+		name     string
+		av       AttachmentValidator
+		errorMsg string
+		wantErr  bool
```
timoreimann, line 34: `wantErr` is not really needed since it derives from `errorMsg != ""`.

driver/mounter_test.go
```diff
@@ -0,0 +1,137 @@
...
+			m := &mounter{
+				log:                 logrus.NewEntry(logrus.New()),
+				attachmentValidator: tt.av,
+			}
+			got, err := m.IsAttached(testSource)
+
+			if tt.wantErr {
+				assert.ErrorContains(t, err, tt.errorMsg)
+			} else {
+				assert.Equal(t, tt.want, got)
```
timoreimann, line 133: nit / personal note: Go stdlib testing [prefers](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures) outputting `got` first, then `want`. I'd match the order here given how thin of a layer gotest.tools is.

timoreimann (commented):
Last round. :)

CHANGELOG.md
llDrLove: I've removed the version header.

driver/mounter.go
llDrLove: I've removed the boolean return and only kept the error. I've refactored the interface method with the new structure, refactored the implementation and the tests.

driver/mounter_test.go
llDrLove: I've removed the wantErr and want properties and now I'm only relying on `errorMsg`.

driver/mounter_test.go
llDrLove: I refactored the test so now it uses the `got` first and `want` after.

timoreimann (approved):
LGTM, thanks a ton Oliver. 👍 

Please squash-merge whenever you're ready.

PR #481, opened 2023-03-14
CON-9080 Remove checkLimit validation in favor of forbidden response when creating a volume 

llDrLove:
Given that the volume limit is to be generalized and raised to 5,000 (which should be more than enough for any [DOKS] customer), showing the actual limit does not provide any value anymore. Thus, we favor handling the http forbidden response from godo for a Forbidden http code (403) with some specific response string than to use the `checkLimit` method.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -181,20 +182,18 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		volumeReq.SnapshotID = snapshotID
 	}
 
-	log.Info("checking volume limit")
-	details, err := d.checkLimit(ctx)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to check volume limit: %s", err)
-	}
-	if details != nil {
-		return nil, status.Errorf(codes.ResourceExhausted,
-			"volume limit (%d) has been reached. Current number of volumes: %d. Please contact support.",
-			details.limit, details.numVolumes)
-	}
-
 	log.WithField("volume_req", volumeReq).Info("creating volume")
-	vol, _, err := d.storage.CreateVolume(ctx, volumeReq)
+	vol, cvResp, err := d.storage.CreateVolume(ctx, volumeReq)
 	if err != nil {
+		if cvResp != nil && cvResp.StatusCode == http.StatusForbidden {
+			var cvErrResp godo.ErrorResponse
+			if err = json.NewDecoder(cvResp.Body).Decode(&cvErrResp); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to decode create volume forbidden response: %s", err)
+			}
+			if strings.Contains(cvErrResp.Message, "capacity limit exceeded") {
+				return nil, status.Errorf(codes.ResourceExhausted, "volume limit has been reached. Please contact support.")
+			}
+		}
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 


driver/controller.go
```diff
@@ -181,20 +182,18 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-	if details != nil {
-		return nil, status.Errorf(codes.ResourceExhausted,
-			"volume limit (%d) has been reached. Current number of volumes: %d. Please contact support.",
-			details.limit, details.numVolumes)
-	}
-
 	log.WithField("volume_req", volumeReq).Info("creating volume")
-	vol, _, err := d.storage.CreateVolume(ctx, volumeReq)
+	vol, cvResp, err := d.storage.CreateVolume(ctx, volumeReq)
 	if err != nil {
```
llDrLove, line 187: @timoreimann , 

Here we have a payload ressembling this : 
```json
{
  "message": "failed to create volume: volume/snapshot capacity limit exceeded",
  "id":"forbidden",
  "request_id": "7c230c90-c7a0-4e79-a40d-b8e3ca003bd4"
}
```

I'm not looking at the `id` field of the `CreateVolume` response for the exact match of `forbidden` since it is not exported by godo. Instead, I'm relying on the http code of the response which is kind of the same check IMO. What do you think ?

timoreimann: Ah, wasn't aware it's not exported. I think your approach makes sense then. 👍

driver/controller.go
```diff
@@ -181,20 +181,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		var errorResponse *godo.ErrorResponse
+		if errors.As(err, &errorResponse) && strings.Contains(err.(*godo.ErrorResponse).Message, "capacity limit exceeded") &&
```
timoreimann, line 188: If we use `errors.As` , then we should also inspect `Message` on `errorResponse` instead of doing another type cast.

Note though how this remark is obsolete now given my next comment.

timoreimann, line 188: I just realized that we do not need to check for and cast into an `ErrorResponse` now that we have discovered that the error ID isn't exported and that we can only check for a portion of the error message: the [`Error()` implementation](https://github.com/digitalocean/godo/blob/2e7120f15645a6399cb37a949bf9a6168b4b6212/godo.go#L476) always includes it, which means it suffices to check `err.Error()` for it.

driver/controller.go
```diff
@@ -181,20 +181,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			cvResp.StatusCode == http.StatusForbidden {
```
timoreimann, line 189: We can add some extra safety by also checking that `cvResp` is not nil.

driver/driver_test.go
```diff
@@ -213,6 +215,10 @@ func (f *fakeStorageDriver) GetVolume(ctx context.Context, id string) (*godo.Vol
 }
 
 func (f *fakeStorageDriver) CreateVolume(ctx context.Context, req *godo.VolumeCreateRequest) (*godo.Volume, *godo.Response, error) {
+	if f.createVolumeErr != nil && f.createVolumeErrResponse != nil {
+		return nil, f.createVolumeErrResponse, f.createVolumeErr
```
timoreimann, line 219: Should we return as soon as one of `f.createVolumeErr` or `f.createVolumeErrResponse` is non-nil (as opposed to requiring both)?

driver/controller.go
llDrLove: Added.

driver/controller.go
llDrLove: True, I refactored the code to check the substring from the basic `err.Error()`.

llDrLove: Won't do based on [comment](https://github.com/digitalocean/csi-digitalocean/pull/481/files#r1137761349) below.

driver/driver_test.go
llDrLove: Sure! It gives more flexibility in the tests. Added.

driver/controller.go
```diff
@@ -181,20 +181,13 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-		return nil, status.Errorf(codes.ResourceExhausted,
-			"volume limit (%d) has been reached. Current number of volumes: %d. Please contact support.",
-			details.limit, details.numVolumes)
-	}
-
 	log.WithField("volume_req", volumeReq).Info("creating volume")
-	vol, _, err := d.storage.CreateVolume(ctx, volumeReq)
+	vol, cvResp, err := d.storage.CreateVolume(ctx, volumeReq)
 	if err != nil {
+		if cvResp != nil && cvResp.Response != nil && cvResp.StatusCode == http.StatusForbidden && strings.Contains(err.Error(), "capacity limit exceeded") {
```
timoreimann, line 187: Small nit / simplification suggestion: I think it's common to check the outer `*godo.Response` for nil-ness only, not the inner `*http.Response`  [since godo always populates them together](https://github.com/digitalocean/godo/blob/2e7120f15645a6399cb37a949bf9a6168b4b6212/godo.go#L378).

driver/controller_test.go
```diff
@@ -284,10 +285,16 @@ func TestControllerExpandVolume(t *testing.T) {
...
+	snapshotId := "snapshotId"
+
 	tests := []struct {
-		name           string
-		listVolumesErr error
-		getSnapshotErr error
+		name                    string
+		listVolumesErr          error
+		getSnapshotErr          error
+		snapchots               map[string]*godo.Snapshot
```
timoreimann, line 294: typo: snapchots -> snapshots

timoreimann (commented):
One tiny suggestion and a typo, otherwise good.

driver/controller.go
llDrLove: Sure, I've removed the `cvResp.Response != nil` check and only kept `cvResp != nil`.

driver/controller_test.go
llDrLove: Nice catch! Fixed.

PR #483, opened 2023-04-03
controller get capacity method added

basit9958:
cc @fatih

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -590,12 +590,28 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 
 // GetCapacity returns the capacity of the storage pool
 func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (*csi.GetCapacityResponse, error) {
-	// TODO(arslan): check if we can provide this information somehow
+	volumes, _, err := d.storage.ListVolumes(ctx, &godo.ListVolumeParams{})
+	if err != nil {
+		return nil, err
+	}
+
+	totalCapacity := int64(0)
+	for _, volume := range volumes {
+		if volume.Region.Slug == d.region {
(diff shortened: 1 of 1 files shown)


driver/controller.go
```diff
@@ -590,12 +590,28 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	totalCapacity := int64(0)
+	for _, volume := range volumes {
+		if volume.Region.Slug == d.region {
+			totalCapacity += int64(volume.SizeGigaBytes) * 1024 * 1024 * 1024
+		}
+	}
+
+	// Create a new response object
+	resp := &csi.GetCapacityResponse{
+		AvailableCapacity: totalCapacity,
```
timoreimann, line 607: AFAIU [the spec for `GetCapacity`](https://github.com/container-storage-interface/spec/blob/b01039c563108173c6743aa1410ec11fde7c24fe/spec.md#getcapacity), the availability capacity is not how much storage capacity is currently in use but how much capacity is available to the user.

I believe this is essentially the volume limit (which we still reference to some extent in the `checkLimit` method that is otherwise unused and was missed for removal in #481) minus the currently used capacity.

timoreimann (commented):
Hey @basit9958 , thanks for your contribution.

I mentioned one relevant point; there are a few others worth calling out, but I'd like to get alignment first on what `GetCapacity` should do in order to make sure we're on the same page.

basit9958:
> Hey @basit9958 , thanks for your contribution.
> 
> I mentioned one relevant point; there are a few others worth calling out, but I'd like to get alignment first on what `GetCapacity` should do in order to make sure we're on the same page.

It should return the total amount of storage capacity that can be allocated for volumes, minus the capacity already in use by existing volumes. I'll update the code tomorrow

driver/controller.go
basit9958: Yes you are correct, We can calculate the used capactity by adding volumesize of volumes whose dropletid is not nil

PR #485, opened 2023-04-18
fixing VolumeSnapshot spec

jkpedo:
If you try and use the existing VolumeSnapshot example this error is returned:

```shell
Error from server (BadRequest): error when creating "snapshot.yaml": VolumeSnapshot in version "v1" cannot be handled as a VolumeSnapshot: strict decoding error: unknown field "spec.snapshotClassName", unknown field "spec.snapshotContentName"
```

### DIFF
--- a/examples/kubernetes/import-snapshot/README.md
+++ b/examples/kubernetes/import-snapshot/README.md
@@ -45,8 +45,9 @@ kind: VolumeSnapshot
 metadata:
   name: snapshot-manual
 spec:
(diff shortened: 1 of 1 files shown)


timoreimann (commented):
👋 thanks for the fix!

Would you mind matching the file from https://github.com/digitalocean/csi-digitalocean/blob/62e330167d9bf1279cbf88b00382c9b12e9243de/examples/kubernetes/import-snapshot/snapshot.yaml?

PR #488, opened 2023-04-28
Disable fail-fast for tests

timoreimann:
If one of the e2e tests fails, the other tests are also canceled. Disable this behavior to minmize the number of complete reruns we have to do.


llDrLove (approved):
LGTM

PR #490, opened 2023-05-03
CON-9337 csi snapshot resizing

llDrLove:
Making sure that the upstream snapshot tests located [here](https://github.com/kubernetes/kubernetes/blob/v1.27.1/test/e2e/storage/testsuites/provisioning.go#L409) is green. 

Adding the functionality of creating a new PVC with a greater capacity from a volume snapshot.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -195,7 +195,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	if snapshot != nil && vol != nil && volumeReq.SnapshotID != "" && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) {
-		log.Info(fmt.Sprintf("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes)))
+		log.Infof("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes))
 		_, _, err = d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", vol.ID, err.Error())
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -188,7 +188,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	vol, cvResp, err := d.storage.CreateVolume(ctx, volumeReq)
 	if err != nil {
 		if cvResp != nil && cvResp.StatusCode == http.StatusForbidden && strings.Contains(err.Error(), "capacity limit exceeded") {
-			return nil, status.Errorf(codes.ResourceExhausted, "volume limit ha s been reached. Please contact support")
+			return nil, status.Errorf(codes.ResourceExhausted, "volume limit has been reached. Please contact support")
 		}
 
 		return nil, status.Error(codes.Internal, err.Error())
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -194,9 +194,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if volumeReq.SnapshotID != "" && snapshot != nil && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) {
-		log.Info("---- hit inside the volume snapshot ")
-		_, _, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
+	if snapshot != nil && vol != nil && volumeReq.SnapshotID != "" && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) {
+		log.Info(fmt.Sprintf("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes)))
+		_, _, err = d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", vol.ID, err.Error())
 		}
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -179,7 +179,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			return nil, err
 		}
 
-		log.WithField("response", resp).Info("snapshot was fetched")
+		log.WithField("snapshot_size_giga_bytes", snapshot.SizeGigaBytes).Info("snapshot size giga bytes")
 		log.WithField("snapshot_id", snapshotID).Info("using snapshot as volume source")
 		volumeReq.SnapshotID = snapshotID
 	}
@@ -196,11 +196,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 
 	if snapshot != nil && vol != nil && volumeReq.SnapshotID != "" && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) {
 		log.Infof("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes))
-		_, _, err = d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
+		_, resp, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", vol.ID, err.Error())
 		}
-		log.Info("volume was resized")
+		log.WithField("response", resp).Info("volume was resized")
 	}
 
 	resp := &csi.CreateVolumeResponse{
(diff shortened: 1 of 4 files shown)


driver/controller.go
```diff
@@ -162,21 +162,24 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		var resp *godo.Response
+		snapshot, resp, err = d.snapshots.Get(ctx, snapshotID)
 		if err != nil {
 			if resp != nil && resp.StatusCode == http.StatusNotFound {
 				return nil, status.Errorf(codes.NotFound, "snapshot %q does not exist", snapshotID)
 			}
 			return nil, err
 		}
 
+		log.WithField("snapshot_size_giga_bytes", snapshot.SizeGigaBytes).Info("snapshot size giga bytes")
```
timoreimann, line 182: I suggest we include the field with the next line where we log a related message about the snapshot anyway.

driver/controller.go
```diff
@@ -191,6 +194,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if snapshot != nil && vol != nil && volumeReq.SnapshotID != "" && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) && int(volumeReq.SizeGigaBytes) > 1 {
```
timoreimann, line 197: Do we need to check `volume.SnapshotID` for emptiness? Shouldn't it be guaranteed to be non-empty when `snapshot` is non-nil?

Also, we don't seem to need the ID further down anyway AFAICT, so it looks to me as if we could ignore it in the if expression.

timoreimann, line 197: Why do we check if the requested size is larger than one GB?

I think we don't need to check for any minimum values because the call in line 96 already looks for boundaries.

driver/controller.go
```diff
@@ -191,6 +194,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		log.Infof("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes))
```
timoreimann, line 198: Grammar: resize -> resized

timoreimann, line 198: Please use `log.WithFields` to show values.

driver/controller.go
```diff
@@ -191,6 +194,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		_, resp, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
```
timoreimann, line 199: Please handle the action: I know resizes finish immediately today, but that's not guaranteed to be the case always.

driver/controller.go
```diff
@@ -191,6 +194,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		log.WithField("response", resp).Info("volume was resized")
```
timoreimann, line 200: Logging the entire response feels wrong (and would probably be hard to read). If we have a specific field we're interested in, let's log that and otherwise drop this line.

If we keep the line, then it likely should only happen after we've handled any error.

driver/controller.go
timoreimann, line 198: Can we include the "why" in the output, namely that the requested volume size is larger than the source snapshot size?

driver/controller.go
llDrLove: IIRC it was because I was getting errors when trying to resize to a 1GB size since the minimal volume is 1GB. If you want to resize , you need to specify a size that is at least greater than 1GB or you will be doing a api/operation for nothing.

timoreimann (commented):
A few small

timoreimann (commented):
.

timoreimann (commented):
A few small comments remaining, mostly optimizations.

timoreimann:
Looks like you have unit tests failing.

driver/controller.go
llDrLove: I tried to add it back but I remembered why I had it removed. I get this error in the logs now : 

May  8 15:18:53 |ERRO| method failed method=/csi.v1.Controller/CreateVolume error=rpc error: code = Internal desc = failed waiting on action ID 1782594914 for volume ID a056bdc3-edb3-11ed-867f-0a58ac14d19d to get resized: timed out waiting for the condition

driver/controller.go
```diff
@@ -179,8 +179,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			return nil, err
 		}
 
-		log.WithField("snapshot_size_giga_bytes", snapshot.SizeGigaBytes).Info("snapshot size giga bytes")
-		log.WithField("snapshot_id", snapshotID).Info("using snapshot as volume source")
+		log.WithField("snapshot_id", snapshotID).
```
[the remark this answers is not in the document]
llDrLove: Thanks! I didn't know that. Refactored both.

driver/controller.go
```diff
@@ -194,13 +195,22 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if snapshot != nil && vol != nil && volumeReq.SnapshotID != "" && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) && int(volumeReq.SizeGigaBytes) > 1 {
-		log.Infof("volume %s will be resize from %d to %d", vol.ID, int(snapshot.SizeGigaBytes), int(volumeReq.SizeGigaBytes))
-		_, resp, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
-		log.WithField("response", resp).Info("volume was resized")
+	if snapshot != nil && int(volumeReq.SizeGigaBytes) > int(snapshot.SizeGigaBytes) && int(volumeReq.SizeGigaBytes) > 1 {
```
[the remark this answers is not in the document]
llDrLove: Fixed.

driver/controller.go
timoreimann: I'd think this must have been a one-off because we also poll the action in the `ControllerExpandVolume` implementation: https://github.com/digitalocean/csi-digitalocean/blob/76367b6a622fe27a517c6ff98c201774040497d2/driver/controller.go#L892-L905

driver/controller.go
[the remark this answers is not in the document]
llDrLove: How would that occur ? Snapshots are created from other volume claims and volume claims create volumes with a integer size between 1 to 10 000 GB. Thus, we are always handling integers here if I understand correctly.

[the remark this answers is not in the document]
timoreimann: Snapshots can also be created outside of Kubernetes. Specifically, given the API allows to specify a float, then it should be possible for a user to also create a snapshot through the API using a float size.

I haven't tested it though, so that's just my reasoning based on type inspections.

driver/controller.go
```diff
@@ -191,6 +197,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if snapshot != nil && volumeReq.SizeGigaBytes > int64(snapshot.SizeGigaBytes) && volumeReq.SizeGigaBytes > 1 {
+		_, _, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
+		log.Info("resizing volume because its requested size is larger than the size of the backing snapshot")
```
timoreimann, line 202: Can we move this above the `Resize` call given it's "announcing" the possibly long-running operation?

driver/controller.go
```diff
@@ -191,6 +197,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", vol.ID, err.Error())
+		}
```
timoreimann, line 205: Where did the action handling go?

driver/controller.go
llDrLove: Moved it up one line.

driver/controller.go
llDrLove: I removed it because it was hanging indefinitely and returning the following error. 
```text
May  8 15:18:53 |ERRO| method failed method=/csi.v1.Controller/CreateVolume error=rpc error: code = Internal desc = failed waiting on action ID 1782594914 for volume ID a056bdc3-edb3-11ed-867f-0a58ac14d19d to get resized: timed out waiting for the condition 
```

timoreimann: We do this elsewhere though: https://github.com/digitalocean/csi-digitalocean/pull/490#discussion_r1187661458

If an action is returned, then I'd strongly believe it's always supposed to be used. If that fails consistently, then we should try to understand why that's the case.

driver/controller.go
```diff
@@ -191,6 +197,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		log.Info("resizing volume because its requested size is larger than the size of the backing snapshot")
+		action, _, err := d.storageActions.Resize(ctx, vol.ID, int(volumeReq.SizeGigaBytes), volumeReq.Region)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "cannot resize volume %s: %s", vol.ID, err.Error())
+		}
+		log = log.WithFields(logrus.Fields{
+			"resized_from": int(snapshot.SizeGigaBytes),
+			"resized_to":   int(volumeReq.SizeGigaBytes),
+		})
+		if action != nil && action.Status != godo.ActionCompleted {
```
timoreimann, line 210: Is either of these ever untrue, i.e., is the action ever missing or its status completed right from the get go?

driver/controller.go
```diff
@@ -191,6 +197,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			log = logWithAction(log, action)
+			log.Info("waiting until volume is resized")
+			waitActionCtx, waitActionCancel := context.WithDeadline(ctx, time.Now().Add(time.Second*15))
```
timoreimann, line 213: IMHO, `context.WithTimeout` would allow for an easier specification of the actual timeout.

Regarding those 15 seconds: do we have a solid understanding by now of the time needed for a resize action to complete?

timoreimann (approved):

PR #502, opened 2023-05-15
CON-9193 Update sidecars containers for v4.6.1

llDrLove:
Updating `deploy/kubernetes/releases/csi-digitalocean-dev` folder to the latest sidecar containers and at the same time create a new v4.6.1 release.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## unreleased
 
+## v4.6.1 - 2023.05.15
```
timoreimann, line 3: Should be 16th by now.

VERSION
```diff
@@ -1 +1 @@
-v4.6.0
+v4.6.1
```
timoreimann, line 1: I think we've never formally defined if we should bump the minor or patch version of the release when we bump just the sidecars. My gut feeling is that when at least one sidecar is bumped on the minor version, then I'd also bump the minor version on our release.

Probably a bit of a grey area, but that's my take.

timoreimann (approved):
LGTM with two comments, though not blocking.

timoreimann (approved):
LGTM with two comments, though not blocking.

VERSION
llDrLove: Same but from this table it seems like everything 1.27 related should be under 4.6.x. Thus, I'll keep it that way for now.

PR #522, opened 2023-07-31
Use godo action status for handling resize operation

timoreimann:
The documented state is now consistently used by the API.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -67,9 +67,6 @@ const (
 	// maxVolumesPerDropletErrorMessage is the error message returned by the DO
 	// API when the per-droplet volume limit would be exceeded.
 	maxVolumesPerDropletErrorMessage = "cannot attach more than 7 volumes to a single Droplet"
-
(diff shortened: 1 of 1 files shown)


gottwald (approved):

driver/controller.go
```diff
@@ -210,7 +207,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			"resized_from": int(snapshot.SizeGigaBytes),
 			"resized_to":   int(volumeReq.SizeGigaBytes),
 		})
-		if action != nil && action.Status != doneActionStatus {
+		if action != nil && action.Status != godo.ActionCompleted {
```
llDrLove, line 210: @timoreimann , is this `action != nil` necessary since we didn't get an error from the `.Resize()` call ? 

Also, here we check for `action != nil` when we don't do the same check on line `1052`. 

It's not really related to your PR but if we could make it consistent it would be better IMO.

PR #571, opened 2024-04-22
update csi for 1.30

dhij:


go.mod
```diff
@@ -1,26 +1,26 @@
 module github.com/digitalocean/csi-digitalocean
 
-go 1.21
+go 1.22.0
```
llDrLove, line 3: @dhij , could this stay at 1.22 ?

dhij: @llDrLove do you mean stay at `1.21` or `1.22`?

dhij: i changed it to `1.22` 🙂

dhij: the CI is now failing: https://github.com/digitalocean/csi-digitalocean/actions/runs/8793921478/job/24132561946?pr=571 weird

timoreimann: You need to run `go mod tidy && go mod vendor` again as the error message implies.

timoreimann: I see that the CI is picking 1.22.0, which is not good because that versions till contains a bunch of CVEs.

I ran across a discussion on Twitter that highlighted this problem with a fix proposal (specifying the version/range in the GHA workflow file) -- see the conversation around this post: https://twitter.com/liggitt/status/1780994064667627993

Could you look into how we can make sure we get to use the latest Go patch release?

dhij: @timoreimann when i run `go mod tidy`, it brings back:
```
go 1.22.0

toolchain go1.22.2
```

And there is no change with `go mod vendor` 🤔

dhij: @timoreimann @llDrLove -

We would have to go with 
```
go 1.22.0
``` 
or 
```
go 1.22
toolchain go1.22.2
```
as suggested [here](https://github.com/golang/go/issues/62278#issuecomment-1693538776).

Without the toolchain, it will complain saying [go: download go1.22 for linux/amd64: toolchain not available](https://github.com/digitalocean/csi-digitalocean/actions/runs/8805591084/job/24168530323#step:6:14).

I also changed the CI to run `make test` instead of `make all` for running the tests. Since Anit raised a question as to whether we should include `check-unused` as part of the "run unit tests" CI job. He also suggested using `go mod tidy -check` for `make check-unused` once https://github.com/golang/go/issues/27005 is fixed.

.github/workflows/test.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Run unit tests
-        run: make all
+        run: make test
```
timoreimann, line 47: I think we do want to run the `check-unused` Makefile target -- otherwise, we can't guarantee that engineers miss to vendor dependencies correctly.

timoreimann (approved):

PR #575, opened 2024-07-30
Make volumes-per-node limit configurable [CON-10973]

timoreimann:
This is to support a future DO block storage volume limit increase.

Adjusting the limit in the driver is to allow the Kubernetes scheduler to work correctly mostly. Specifically, it won't have an influence on what the DO storage backend permits. The two need to be synchronized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## unreleased
 
+* Make volumes-per-node limit configurable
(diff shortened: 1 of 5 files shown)


m3co-code (approved):
LGTM, not even sth wort a nit-pick :3  

Awesome to see such a short cycle time from discussion to implementation.

PR #591, opened 2024-08-16
Prepare release v4.12.0

d-honeybadger:
Got wondering for a bit if this should be a major release or not, due to the breaking change in exernal-snapshotter we pulled in, but ultimately decided it's not a breaking change for csi-digitalocean

- there's nothing breaking in the digitalocean/do-csi-plugin:v4.12.0 image/binary itself; users who just pull in the new image will be fine
- if someone grabs all the crds and example deployments from our repo, they'll be fine too cause the crds and the webhook will be updated in conjunction

UPD: I'm probably overthinking it. The 4.12.0 version is for do-csi-plugin, the validating webhook configuration is just here so that we can e2e testing and for convenience


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 ## unreleased
 
+## v4.12.0 - 2024.08.16
+
+* Update CSI driver for Kubernetes 1.31
+  Update CSI sidecars to latest
+  Update Go to 1.23
+  Update other go.mod dependencies
+  Update VolumeSnapshot, VolumeSnapshotContents, and VolumeSnapshotClasses CRDs to pull in CEL validation rules new in external-snapshotter v8 release
+  Remove VolumeSnapshot and VolumeSnapshotContets from the snapshot validating webhook configuration as those are now validated via CEL rules on their CRDs
```
dhij, line 10: nit: typo with VolumeSnapshotContets

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 ## unreleased
 
+## v4.12.0 - 2024.08.16
+
+* Update CSI driver for Kubernetes 1.31
+  Update CSI sidecars to latest
```
dhij, line 6: i think it will be cleaner to have them in bullet points `*` like done below and also link the relevant PRs 🙂

d-honeybadger: They're all in the same PR though

CHANGELOG.md
d-honeybadger: fixed - TY!

CHANGELOG.md
d-honeybadger: Oh it's all going in the same paragraph when rendered. Let me find something better

d-honeybadger: Added newlines so that in the rendered .md those are kept on separate lines
Still a single PR though

dhij (approved):
LGTM!

PR #631, opened 2025-03-17
update external-resizer

d-honeybadger:
There was an issue with external-resizer v1.13 failing the `Dynamic PV (default fs)(allowExpansion)` test from upstream k8s suite.

The bug was actually in the k8s e2e tests: https://github.com/kubernetes/kubernetes/pull/129085
The upstream fix was merged and backported to k8s 1.30 - 1.32, so here I'm updating the e2e test binaries and removing 1.29 version since a) the fix wasn't backported to 1.29 and b) it's already removed in doks


m3co-code (approved):
LGTM :)

PR #633, opened 2025-03-21
Fix mounting of XFS volumes restored from snapshots

d-honeybadger:
Applying essentially the same workaround as other providers, e.g. https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/788


m3co-code (approved):
Great work!

PR #646, opened 2026-04-19
Fixed the max volume mount default from 7 to 15

jasondcamp:
I ran into an issue with the maximum volume mounts hitting the limit of 7. According to https://docs.digitalocean.com/products/volumes/details/limits/ the limit is raised to 15 now.  

There was a const that defined the value, as well as a command line flag, i consolidated it down to a single source of truth in the const, and also updated the case to match other const patterns.

I also fixed a typo in some tests that said test_enabed, changed to test_enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## unreleased
 
+* Adjusted max volume mounts from 7 to 15
+
 ## v4.17.0 - 2026.05.06
 
 * Update Go to 1.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## unreleased
 
-* Adjusted max volume mounts from 7 to 15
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## unreleased
 
+* Adjusted max volume mounts from 7 to 15
```
varshavaradarajan:
```suggestion
* Adjusted default max volume limit per node from 7 to 15.
```

varshavaradarajan (changes requested):

varshavaradarajan:
@jasondcamp - thank you for the PR! Apologies on the delay in review. We've had to focus on other roadmap priorities.

jasondcamp:
No worries at all, I appreciate you taking the time to implement this!

varshavaradarajan (approved):

PR #654, opened 2026-08-13
Remove hardcoded maximum volume size check

dhij:
## Summary

- Removes the hardcoded 16Ti `maximumVolumeSizeInBytes` check in `extractStorage`, so oversized `CreateVolume`/`ControllerExpandVolume` requests are validated by the DigitalOcean API's own per-volume size limit instead of a value duplicated in the driver.
- Maps the API's resulting error (`codes.InvalidArgument` → HTTP 422, message "invalid size specified") to `codes.OutOfRange`, matching CSI semantics for a bad capacity request. Verified against the live API: an oversized `CreateVolume` call returns `422 {"message":"failed to create volume: invalid size specified", ...}`.
- Also broadens the existing capacity-limit-exceeded check to match on HTTP 403 or 429, since the underlying gRPC→HTTP mapping table returns 429 for `codes.ResourceExhausted` while the current code only checked for 403.
- Updated to reference [new godo constants](https://github.com/nayanjd-do/godo/blob/c602a1c4015133a6212d272364efac6ee9be1413/storage.go#L20-L37) for errors

## Test plan

- [x] `go test ./driver/...` passes (new cases cover the 422/`OutOfRange` mapping for both `CreateVolume` and `ControllerExpandVolume`, and the 403/429 capacity-limit cases)
- [x] `go build ./...` / `go vet ./...` clean (pre-existing vet warning unrelated to this change)


varshavaradarajan (approved):

dhij:
this PR adds `hostNetwork:true` to our csi controller in e2e manifest (`test/kubernetes/deploy/kustomization.yaml`). because:

during e2e, our test harness (`test/kubernetes/deploy/deploy.sh`) doesn't replicate DOKS's real CPC/data-plane split at all. It just takes the kubeconfig of the one customer cluster we created and kubectl applys the entire driver, controller included, directly onto it. And since the controller relies on the metadata service to retrieve the region, it gets blocked by cilium's egress policy that blocks traffic to our IMDS.

